PR #39, opened 2018-02-15
Fix anchor links to headings

zcei:
Since emoji headings were introduced in https://github.com/oclif/oclif/commit/fb7765cf6e6aee9be110eab94e596cdc85c323dc, the links to headings were broken at a few points.

I took the chance to validate all the links in the readme (via [remark](https://remark.js.org/)) and fix them.

The only outlier is "Other Command Options", as the cog seems to have a variation selector in front, which makes GitHub treat the anchor as `#%EF%B8%8F-other-command-options`. We could potentially hardcore the link to this, but maybe we can find a different solution to this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,10 @@ This is a framework for building CLIs in Node.js. This framework was built out o
 # ✨ Features
 
 * **Flag/Argument parsing** - No CLI framework would be complete without a flag parser. We've built a custom one from years of experimentation that we feel consistently handles user input flexible enough for the user to be able to easily use the CLI in ways they expect, but without comprisiming strictness guarantees to the developer.
-* **CLI Generator** - Run a single command to scaffold out a fully functional CLI and get started quickly. See [Usage](#usage) below.
+* **CLI Generator** - Run a single command to scaffold out a fully functional CLI and get started quickly. See [Usage](#-usage) below.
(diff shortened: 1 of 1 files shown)


jdxcode:
the command ones are autogenerated from https://github.com/oclif/dev-cli/blob/master/src/commands/readme.ts so they'll just get overwritten the next time the package is published. I'll look into fixing that

zcei:
If I'm not mistaken one would need to "slugify" the `commandUsage` value here instead of using the command id:
https://github.com/oclif/dev-cli/blob/master/src/commands/readme.ts#L139

jdxcode:
already done!

jdxcode:
there is probably a better slugify function I could use, but this works for now

zcei:
I like the idea of auto-generating the readme from the actual usage instructions of the CLI.
Thinking about generating an [`mdast`](https://github.com/syntax-tree/mdast) compatible notation, that could then be automatically validated and stringified in a consistent style.

jdxcode:
I like that idea. My original idea was to have multiple output formats (console help, markdown, or man pages), but that was pretty hard to do actually.

This isn't documented yet but there is an alternate markdown output where you can have it list out each topic: https://github.com/heroku/cli/tree/oclif#command-topics Which is useful for very large CLIs like heroku.

nsamsami:
@zcei thanks for checking out oclif! I'm a product manager at Heroku where we've been developing the project. As we're just getting started, I'd be really interested in your feedback and learning more about your use case. Feel free to respond here or drop us a note at heroku-cli-squad@salesforce.com

PR #43, opened 2018-03-13
fix: return a promise from yeoman callback style so it actually awaits

amphro:
Yeomans env.run doesn't seem to return a promise, but takes a callback.

### DIFF
--- a/src/command_base.ts
+++ b/src/command_base.ts
@@ -10,6 +10,11 @@ export default abstract class CommandBase extends Command {
       `oclif:${type}`
     )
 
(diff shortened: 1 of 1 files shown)


jdxcode:
@amphro I would do this on your behalf but I don't think I can. Can you rename the commit message to "fix: return a promise from yomen callback style so it actually awaits". This way it will go into the changelog correctly

amphro:
@jdxcode doesn't seem to like that one either

jdxcode:
it's fine, I was toying with git cop to make it clearer but it doesn't really work right

PR #53, opened 2018-03-20
Update README.md

nsamsami:
Added link to getting started tutorial


jdxcode:
@nsamsami can we place this more towards the top? Also, if you can find an emoji that you think fits for it then it'll match the other headers a bit better

Issue #51, opened 2018-03-16
make tslint config optional


jdxcode:
we could possibly offer an alternate tslint (with semicolons, airbnb, etc) but that can come later. For now, we should make it so tslint is only installed if desired like with mocha and typescript.

Changes needed:

- [ ] add option to disable tslint
- [ ] only add tslint.json when tslint is specified
- [ ] only add tslint to package.json when tslint is specified

PR #63, opened 2018-03-21
Add TSLint optional

gfvcastro:
Closes https://github.com/oclif/oclif/issues/51

### DIFF
--- a/src/generators/app.ts
+++ b/src/generators/app.ts
@@ -33,6 +33,7 @@ class App extends Generator {
     mocha: boolean
     'semantic-release': boolean
     typescript: boolean
+    tslint: boolean
   }
   args!: {[k: string]: string}
   type: 'single' | 'multi' | 'plugin' | 'base'
@@ -52,12 +53,14 @@ class App extends Generator {
     options: {
       mocha: boolean
       typescript: boolean
+      tslint: boolean
       'semantic-release': boolean
     }
   }
   mocha!: boolean
   semantic_release!: boolean
   ts!: boolean
+  tslint!: boolean
   get _ext() { return this.ts ? 'ts' : 'js' }
   get _bin() {
     let bin = this.pjson.oclif && (this.pjson.oclif.bin || this.pjson.oclif.dirname) || this.pjson.name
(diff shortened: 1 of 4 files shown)


src/generators/app.ts
```diff
@@ -233,9 +239,9 @@ class App extends Generator {
     this.pjson.repository = this.answers.github ? `${this.answers.github.user}/${this.answers.github.repo}` : defaults.repository
     this.pjson.scripts.posttest = 'yarn run lint'
     // this.pjson.scripts.precommit = 'yarn run lint'
-    if (this.ts && this.mocha) {
+    if (this.ts && this.tslint && this.mocha) {
       this.pjson.scripts.lint = 'tsc -p test --noEmit && tslint -p test -t stylish'
```
jdxcode: if we don't select tslint, we should still run `tsc -p test`. The way this is written right now it would fall back to `eslint` which doesn't come with typescript setups

gfvcastro: Oops, good catch!

src/generators/app.ts
```diff
@@ -233,10 +239,12 @@ class App extends Generator {
...
-    } else if (this.ts) {
+    } else if (this.ts && this.tslint) {
       this.pjson.scripts.lint = 'tsc -p . --noEmit && tslint -p . -t stylish'
+    } else if (this.ts && this.mocha) {
+      this.pjson.scripts.lint = 'tsc -p test --noEmit'
```
jdxcode: hmm, we use `-p test` when use mocha and `-p .` when we don't have mocha (this way it builds the mocha files only when doing the linting)

Maybe we could rewrite this like the following to clean it up:

```js
if (this.ts) {
  const tsProject = this.mocha ? 'test' : '.'
  this.pjson.scripts.lint = `tsc -p ${tsProject}`
  if (this.tslint) this.pjson.scripts.lint += `&& tslint -p ${tsProject}`
}
```

I _think_ that's right. What do you think?

src/generators/app.ts
```diff
@@ -420,9 +430,14 @@ class App extends Generator {
         // '@types/supports-color',
         'typescript',
         'ts-node@5',
-        '@oclif/tslint',
-        'tslint',
+        'tslib',
```
jdxcode: 👍

PR #74, opened 2018-03-22
getting started with oclif without npx guide

devMYC:
got confused, is this fix correct?


jdxcode:
That part had a couple of issues. It had some extra characters that must've been adding during a copy+paste, but it also was confusing because it was trying to show 2 things at once. First that you can install without npx, also that you can run the generator without specifying the project name to use the current directory.

I think the second part is less important and I'm going to move that into the commands documentation.

I made a couple commits to your PR to fix it up a bit more. If you wouldn't mind signing the CLA (legal things we have to do) I'd be happy to merge. Thanks for the contribution!

Issue #122, opened 2018-05-29
Use of `rm` command in npm scripts is not cross-platform


mattmazzola:
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------
Request a feature. The npm scripts should be cross-platform compatible. E.g. using `rimraf` instead of `rm` directly

What is the current behavior?
-----------------------------
steps to reproduce:
```
npx oclif multi test-01
cd test-01
npm run publish
```
This implicitly runs the prepack commands which use `rm` and fails because `rm` is not found.

What is the expected behavior?
------------------------------
npm scripts should run successfully after fresh new app with no modifications

```
    "@oclif/command": "1.4.30",
    "@oclif/config": "1.6.19",
    "@oclif/plugin-help": "1.2.11",
```

Windows 10 -  10.0.17134 Build 17134

jdxcode:
duplicate of https://github.com/oclif/oclif/issues/56

PR #123, opened 2018-05-29
Add rimraf to devDependencies for single and multi projects and use it in npm scripts

mattmazzola:
This was one way to address cross-platform script issue although there might be better alternative.

I'm not sure if the 'plugin' project type also needs the devDependency

Closes: #122


mattmazzola:
I added the platform checks for win32 although I feel like the attempt to reduce "complexity" from unix users just absorbed the complexity into maintaining this file.  If using rimraf added complexity that would be arguable, but it was drop in replacement. Adding the checks for win32 make it more complicated and brittle.

> this wouldn't work if the platform was initially on unix then changed to windows

I think this is more likely to occur than you would think.  Imagine a team of people working to build a CLI, it was scaffolded by unix user so the package json has `rm -rf` in it and then later cloned by another team member who uses windows. It will error on his machine and they'll have to re-perform the switch to rimraf and understand all the places it's necessary which adds barrier to usage for something trivial to take care of for them.

Issue #68, opened 2018-03-22
make circleci optional


jdxcode:

jdxcode:
yes, there should be an option to check to add the 2 circleci files

Horaddrim:
Okay, I'll try to PR this :smile: and, first of all! Congratulations for your job @jdxcode I really like people who love CLIs :D

aequasi:
Would be nice if greenkeeper also would use npm if yarn isnt available.

CallumDenby:
Implemented in #137 can close now @jdxcode

PR #137, opened 2018-07-09
feat: make circleci optional

CallumDenby:
Addresses PR #68 

CircleCI is now an option in the CLI with a default to true, the same as Mocha.


jdxcode:
Looks great! Thanks so much!

PR #151, opened 2018-08-19
fix: remove npm-shrinkwrap from multi

Aghassi:
Since this is generating new CLIs for Yarn and NPM, both of which leverage lock files, seems like there shouldn't be a need to rm and bundle shrink wraps....


src/generators/app.ts
```diff
@@ -288,10 +288,9 @@ class App extends Generator {
     }
     if (['plugin', 'multi'].includes(this.type)) {
       this.pjson.scripts.prepack = nps.series(this.pjson.scripts.prepare, 'oclif-dev manifest', 'oclif-dev readme', 'npm shrinkwrap')
```
Aghassi: Probably worth removing this too.

jdxcode:
lock files are not used when installing packages. That's why we include the shrinkwrap

PR #152, opened 2018-08-19
fix: update eslint to latest major in generator

Aghassi:
Debating whether this should be set to `latest` or just have to manually update. Eslint is now on major version 5, so this should reflect that...

### DIFF
--- a/src/generators/app.ts
+++ b/src/generators/app.ts
@@ -471,7 +471,7 @@ class App extends Generator {
       }
     } else {
       devDependencies.push(
-        'eslint@^4',
+        'eslint@^5',
         'eslint-config-oclif@^1',
       )
     }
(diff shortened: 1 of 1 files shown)


jdxcode:
moving it to latest would cause it to break if there were any incompatibilities

Aghassi:
@jdxcode thanks for taking in this PR. Follow up question. I noticed that a number of dependencies are constructed via JS instead of having a template package.json. If this repo were to leverage something like http://renovatebot.com, this sort of thing could be avoided as it could monitor the package.json files and open PRs against them when new versions come out. Then, tests could ensure the generator still builds and works fine. Is there a reason this sort of system isn't in place? Is it worth me making an issue and maybe taking up in my free time?

jdxcode:
I don't think it would be possible to do something like that at a meta level such as this. I'm pretty happy with how it works today. After the generator runs you can always just update packages.

PR #217, opened 2019-03-14
Added A Feature Requested

### LINKED ISSUE #174: Feature request: don't assume every project will be stored on Github
A feature.

What is the current behavior?
-----------------------------
When creating a new app (for example: `npx oclif single testapp`) I am prompted for Github owner and repository name.

I should be asked which service I am using before being asked for the owner username; at the least, I should be given a choice of Github, Bitbucket, GitLab, as well as an option to specify a private server (Github Enterprise, GitLab-CE etc — I guess this would require the base URL to the instance as a followup question).

A "none of the above" option which leaves `bugs`, `homepage` and `repository` blank might be useful, too.

In a totally ideal world, I would be able to set my default as an environment variable, or in a .rc file somewhere, so the defaults are correct each time.

The `repository` entry in the generated `package.json` should be explicit in specifying the server:

Server|repository
------|----------
Github|github:unikitty37/testapp
Github Enterprise|https://github.example.com/unikitty37/testapp
Bitbucket|bitbucket:unikitty37/testapp
GitLab|gitlab:unikitty37/testapp
GitLab-CE|https://gitlab.example.com/unikitty37/testapp

The `bugs` and `homepage` entries should default to the correct format for the services concerned (which should be a case of just replacing `github.com` with the service root — GitLab projects in a group will have an extra layer between username and app name, but that's probably more complication than needed, at least for an initial release…)

Currently, this can be worked around by manually editing `package.json` after the fact, but that's a fair bit of extra work for people not using Github. Github could remain the default service selection, meaning Github users have only one extra press of the Enter key in the app generation workflow.

I might be able to look at implementing this at some point in the future, but I'm a bit snowed under at the moment; would there be interest in adding this?

oclif 1.12.1; Mac OS X 10.14 (18A391)

aggmoulik:
Resolved #174

### DIFF
--- a/src/generators/app.ts
+++ b/src/generators/app.ts
@@ -73,6 +73,13 @@ class App extends Generator {
       appveyor: boolean
       codecov: boolean
     }
+    service: {
+      github: boolean
+      githubCE: boolean
+      bitbucket: boolean
+      gitlab: boolean
+      gitlabCE: boolean
+    }
   }
   mocha!: boolean
   circleci!: boolean
@@ -82,6 +89,11 @@ class App extends Generator {
   tslint!: boolean
   eslint!: boolean
   yarn!: boolean
+  github!: boolean
+  githubCE!: boolean
+  bitbucket!: boolean
+  gitlab!: boolean
+  gitlabCE!: boolean
   get _ext() { return this.ts ? 'ts' : 'js' }
   get _bin() {
     let bin = this.pjson.oclif && (this.pjson.oclif.bin || this.pjson.oclif.dirname) || this.pjson.name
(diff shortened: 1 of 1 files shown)


moulikcipherX:
@RasPhilCo @jdxcode

moulikcipherX:
@jdxcode @RasPhilCo
Please review this PR sir..

moulikcipherX:
@jdxcode

dotnetCarpenter:
This actually looks like a good PR. Even though its explanation is horrible. The title should probably be something like _"Added support for githubCE, bitbucket, gitlab and gitlabCE"_.

That said, the old code is commented out instead of removed and [the git service section](https://github.com/oclif/oclif/pull/217/files#diff-e8e3ac8726d6037e00a7fe303119ff69L300) has the same code for github/githubCE and gitlab/gitlabCE, which makes me think that this is untested code. I could be wrong though.

RasPhilCo:
Closing due to age. If you think this is in error, please reply.

PR #220, opened 2019-03-22
Modify README.md

### LINKED ISSUE #140: Node package recommendation question
Documentation "bug" possibly.

What is the current behavior?
-----------------------------

It is advised that one can simply add the node package to the deps, and users will be running the latest node.  This doesn't seem totally correct though.

> You can add the node package to your CLI to ensure users are on Node 8.

This will only be true when your cli is run from a script field or a tool like npx, not when installed globally or used through the shebang bin file.

The current wording makes it sound like a generalized solution when it isn't.  Perhaps we should clarify where this works and doesn't work?

sourabh1031:
fixes #140

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ $ heroku info --app=<tab><tab> # will complete with all the Heroku apps a user h
 
 # 📌 Requirements
 
-Only Node 8+ is supported. Node 6 will reach end-of-life April 2019. At that point we will continue to support the current LTS version of node. You can add the [node](https://www.npmjs.com/package/node) package to your CLI to ensure users are on Node 8.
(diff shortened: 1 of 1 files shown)


sourabh1031:
@jdxcode @RasPhilCo please review

sourabh1031:
and one more thing i got a nice article on the web
 https://medium.com/the-z/getting-started-with-oclif-by-creating-a-todo-cli-app-b3a2649adbcf . 
This is best to be included in the documentation itself. It will certainly help newbies to get started with oclif.

sourabh1031:
@RasPhilCo @jdxcode please review

RasPhilCo:
Closing due to age. If you think this is in error, please reply.

PR #232, opened 2019-05-06
fix: remove buildinfo on composite ts builds

rdaniels6813:
If you are building locally using `npm version` then `npm publish` the lib folder doesn't show up in the published package.
This occurs because we are doing `rm -rf lib` on prepack in addition to having `"composite": true` in tsconfig.  The lib folder won't rebuild locally until you also remove tsconfig.tsbuildinfo file, so having this be a part of the 'prepack' step will solve this issue locally.

I have a feeling that CI builds didn't see this since they are likely being run against a fresh repo.

### DIFF
--- a/src/generators/app.ts
+++ b/src/generators/app.ts
@@ -310,7 +310,7 @@ class App extends Generator {
       this.pjson.scripts.test = 'echo NO TESTS'
     }
     if (this.ts) {
-      this.pjson.scripts.prepack = nps.series(`${rmrf} lib`, 'tsc -b')
+      this.pjson.scripts.prepack = nps.series(`${rmrf} lib`, `${rmf} tsconfig.tsbuildinfo`, 'tsc -b')
     }
     if (['plugin', 'multi'].includes(this.type)) {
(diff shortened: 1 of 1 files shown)


jdxcode:
that's not the standard location for [tsbuildinfo](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html) though. It should already be inside `./lib`

rdaniels6813:
When I run `npm run prepack` locally, it creates it at the root of the project.  If I remove the 'rootDir' property from tsconfig.json it works as you described.  Either way, I am seeing this with a fresh project created with `oclif multi`.  This might be undocumented or buggy behaviour from the typescript end, but would still be nice if this build process caught/fixed it.

Yeti-or:
any progress here?

RasPhilCo:
I don't think this is needed needed now since https://github.com/oclif/oclif/commit/25fde1c4ab6a5c36b8f8300831ba9cd5f8b25e41

rdaniels6813:
@RasPhilCo Correct, closing this now.

Issue #240, opened 2019-05-28
Oclif command scaffold has outdated license


joshwlewis:
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------
*bug*

What is the current behavior?
-----------------------------
`npx oclif plugin myplugin` yields a project with a 2018 license.

```bash
$ npx --version
6.4.1
$ node --version
v10.15.3
$ npx oclif plugin myplugin
...
$ cat LICENSE | grep Copyright
Copyright (c) 2018 @joshwlewis
```

What is the expected behavior?
------------------------------
`npx oclif plugin myplugin` yields a project with a 2019 license.

jdxcode:
[The issue is that it's using windows line endings in the travis file](https://github.com/janpio/oclif-ci-demo/blob/master/.travis.yml#L8)

RasPhilCo:
Fixed via #241

PR #241, opened 2019-05-28
Update license template copyright year

joshwlewis:
There might be a better fix for #240 that makes the year dynamic, but this may work for a quick and temporary solution.


joshwlewis:
Closed and reopened to trigger CLA status.

PR #327, opened 2020-02-17
fix: short circuit with nullish coalescing operator

nicholasjpaterno:


RasPhilCo:
Thanks! Can ship this after singing the CLA.

Issue #423, opened 2020-08-20
eslint error when run `npm test`


imnemo:
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------
bug

What is the current behavior?
-----------------------------
When I run `npm test`, I see an `eslint` error info shown in the image below...
![image](https://user-images.githubusercontent.com/1251690/90764067-53a1f800-e31a-11ea-8f75-8636dcd20c23.png)

It is something like...
```bash
/Users/nemo/code/nodelab/clilab/ocliff/hello/src/commands/hello.ts
  14:16  error  Missing trailing comma  comma-dangle

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

npm ERR! Test failed.  See above for more details.
```

If the current behavior is a bug, please provide the steps to reproduce.
```bash
oclif multi hello
cd hello
npm test
```

What is the expected behavior?
------------------------------
No any error or warning info!

Please mention your oclif and OS version.
---
OS version: v10.15.5 (macOS Catalina)
oclif version: v1.16.1

RasPhilCo:
I'm unable to reproduce this. Do you have another eslint config loading rules (global)?

RasPhilCo:
Cleaning up - if you think this issue is still valid, reopen it. Thanks!

PR #424, opened 2020-08-20
fix: eslint comma-dangle error when run _npm run_ (#423)

imnemo:
fix issue #423


RasPhilCo:
Thanks for the PR! However, this is a rule we want to keep.

PR #593, opened 2021-05-27
Sm/index version limits

mshanemc:
@W-9326244@

Add indexVersionLimit to pjson s3 props
If it exists, the index files should only have the last [indexVersionLimit] version


amphro (approved):

amphro:
@mshanemc you have lint errors.

PR #607, opened 2021-07-01
fix: use @oclif/core for readme/manifest

amphro:


src/commands/manifest.ts
```diff
@@ -16,15 +15,15 @@ export default class Manifest extends Command {
...
     const root = path.resolve(args.path)
-    let plugin = new Config.Plugin({root, type: 'core', ignoreManifest: true, errorOnManifestCreate: true})
+    let plugin = new Plugin({root, type: 'core', ignoreManifest: true, errorOnManifestCreate: true})
     if (!plugin) throw new Error('plugin not found')
     await plugin.load()
     if (!plugin.valid) {
       const p = require.resolve('@oclif/plugin-legacy', {paths: [process.cwd()]})
       const {PluginLegacy} = require(p)
-      delete plugin.name
+      // delete plugin.name
```
mdonnalley, line 26: Does this need to be removed? uncommented?

mdonnalley (approved):

src/commands/manifest.ts
amphro: I'll remove. Not sure if it is needed but it didn't exist in core and didn't seem to matter.

PR #672, opened 2021-09-21
fix: update package.json deps & CI for oclif@v2

RodEsp:
* Updates the deps in the `package.json` to use the newer versions of packages that now use `@oclif/core`.
* Updates the `scripts` section accordingly

Requires https://github.com/oclif/eslint-config-oclif-typescript/pull/30 and https://github.com/oclif/eslint-config-oclif/pull/85/


peternhale (approved):

RasPhilCo:
Curious if the updated oclif eslint package fixes this

RodEsp:
@RasPhilCo it does! That's why I did it. Just gotta update this too.

RodEsp:
Unfortunately it adds a bunch of other eslint issues, although these are actually not bugs and _should_ be fixable!

package.json
```diff
@@ -29,9 +28,7 @@
     "yosay": "^2.0.2"
   },
   "devDependencies": {
-    "@oclif/dev-cli": "^1.21.0",
-    "@oclif/plugin-legacy": "^1.1.4",
```
RodEsp, line 33: Heya @RasPhilCo curious on your opinion on removing `plugin-legacy`. I'm not _really_ sure if it's _needed_ or not but thought it would be a good opportunity to remove with oclif/core stuff?

.circleci/config.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - restore_cache: *restore_cache
       - attach_workspace: {at: node_modules}
       - run: .circleci/setup_git
-      - run: yarn exec nps test
```
RasPhilCo, line 44: nps runs the `generate` cmd which creates a cli and runs it. all for removing nps but we're losing coverage here we need to make up some other way.

package.json
RasPhilCo: Cool!

Maybe this PR should be a few separate ones? test changes, linting changes, package changes

.eslintrc
```diff
@@ -3,7 +3,10 @@
     "oclif",
     "oclif-typescript"
   ],
-  "rules": {
-    "unicorn/no-abusive-eslint-disable": 0
-  }
+  "overrides": [{
+    "files": [ "./test/**/*" ],
```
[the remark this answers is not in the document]
RodEsp: It does lint them, it just overrides the `"node/no-extraneous-import"` rule for test files because the fixtures have their own `package.json`s that have no dependencies so eslint complains that you shouldn't use dependencies that belong to some other package.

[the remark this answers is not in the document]
RasPhilCo: Oh I see/read!

PR #673, opened 2021-09-21
build: use bin/dev to create manifest & readme

RasPhilCo:
Take a look at the release job, it is failing by trying to use `bin/dev` which only works if the lib is build. This uses bin/dev to create the manifest and readme for releasing.


RodEsp (approved):

MMCKNurse:
How do you re-create the manifest after running prepack? Sorry i'm new.

PR #674, opened 2021-09-25
ci: use node:lastest

RasPhilCo:


RodEsp:
@RasPhilCo what is the point of this job? It does something with `nyc` but I don't understand for what purpose.

RasPhilCo:
you can merge this in and pass your tests

RodEsp:
@RasPhilCo yeah but I'm thinking if it doesn't serve a purpose anymore we can just get rid of it?

There's already another testing job that doesn't rely on that `nsis` docker image.

PR #696, opened 2021-10-28
Allow specifying an additional cli for installers

RodEsp:
## What does this PR do?
* Adds a hidden flag, `--additional-cli`, to the `oclif pack:win` and `oclif pack:macos` commands that will make an additional Oclif CLI available on the user's machine. In order for this to work correctly the following needs to be true:
    * There needs to be an executable named after the additional CLI in the current CLI's `./bin` directory (a `.cmd` file for Windows and a shell script for Mac).
    * The additional CLI needs to be added at the root of the current CLI's tarball in a folder named after itself.
        * See [`include-sf.js`](https://github.com/salesforcecli/sfdx-cli/blob/main/scripts/include-sf.js) for an example of a [pretarball script](https://github.com/oclif/oclif/blob/master/src/tarballs/build.ts#L70) that does this.
* Fixes several bugs with how tarballs and installers were being uploaded. The tarball config version and package version were different and weren't being used appropriately when dealing with `prerelease` versions.
* Updates CI to use non-outdated versions of node.
* Allows running npm/yarn scripts on Windows with `shx`.

## Testing Instructions
1. Uninstall all versions of the CLI on your machine using the instructions [here](https://developer.salesforce.com/docs/atlas.en-us.234.0.sfdx_setup.meta/sfdx_setup/sfdx_setup_uninstall.htm).
1. Download the testing installer using the AWS CLI:
    1. Mac: 
    ```
    aws s3 cp s3://dfc-data-production/media/salesforce-cli/sfdx/versions/7.126.0-dev.8/47dc692/sfdx-v7.126.0-dev.8-47dc692.pkg ./
    ```
    1. Windows:
    ```
    aws s3 cp s3://dfc-data-production/media/salesforce-cli/sfdx/versions/7.126.0-dev.8/47dc692/sfdx-v7.126.0-dev.8-47dc692-x64.exe ./
    ```
    1. Review [the PR that generated those installers](https://github.com/salesforcecli/sfdx-cli/pull/329).
1. Run the installer.
1. Check that both `sfdx` and `sf` are available (you might need to restart your terminal for env vars to take effect).
    1. Versions should be:
        1. sfdx - `7.126.0-dev.8`
        1. sf - `1.2.0`
        1. Using node `16.13.0`
1. Run `sfdx update dev`
    1. Check CLIs are available as above.
1. Run `sfdx update stable`
    1. Check CLIs are available as above (`sf` will still be available because files aren't deleted with `update` just overwritten).
1. Run `sfdx update dev` again.
    1. Check CLIs are available as above.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ jobs:
       - attach_workspace: {at: node_modules}
       - run: yarn
       - run: rm -rf lib && npx tsc --noEmit && rm -fr lib
-  test:
+  node-latest-multi:
     <<: *lint
     parameters:
       image:
@@ -93,6 +93,7 @@ workflows:
           requires: [lint, compile-typescript]
       - cache: {requires: [release-management/release-package] }
       - release-management/release-package:
+          context: org-global
           github-release: true
           prerelease: true
           tag: core-v1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,9 +15,8 @@ jobs:
       - checkout
       - restore_cache: &restore_cache
           keys:
-            - v5-{{checksum ".circleci/config.yml"}}-{{ checksum "yarn.lock"}}
-            - v5-{{checksum ".circleci/config.yml"}}
-            - v5
+            - cache-{{checksum ".circleci/config.yml"}}-{{ checksum "yarn.lock"}}
+            - cache-{{checksum ".circleci/config.yml"}}
       - run: yarn add -D mocha-junit-reporter@1
       - run: yarn exec nps lint
       - persist_to_workspace: {root: node_modules,  paths: ["*"]}
@@ -54,8 +53,8 @@ jobs:
       - checkout
       - restore_cache: &restore_cache
           keys:
-            - v6-yarn-{{checksum ".circleci/config.yml"}}-{{ checksum "yarn.lock"}}
-            - v6-yarn-{{checksum ".circleci/config.yml"}}
+            - cache-{{checksum ".circleci/config.yml"}}-{{ checksum "yarn.lock"}}
+            - cache-{{checksum ".circleci/config.yml"}}
       - run: 'echo "Node: `node --version`"'
       - run: yarn
       - run: .circleci/loadgpg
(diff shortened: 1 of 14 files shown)


src/commands/pack/macos.ts
```diff
@@ -103,6 +108,8 @@ export default class PackMacos extends Command {
 
   static flags = {
     root: flags.string({char: 'r', description: 'path to oclif CLI root', default: '.', required: true}),
+    'additional-cli': flags.string({description: `an Oclif CLI other than the one listed in config.bin that should be made available to the user
```
mdonnalley, line 111: This feels like a flag that should be hidden - what do you think?

mdonnalley:
@RodEsp I'm assuming that you've targeted `master` because sfdx-cli is still using v1? I think I'd prefer for this to go into `main` and then update sfdx-cli to use v2 - unless there's some reason that sfdx-cli can't use v2

RodEsp:
@mdonnalley sounds good. I can test generating the `sfdx-cli` installers using an oclif@v2 version of this.

RodEsp:
@peternhale @mdonnalley this is fully ready. I don't expect any more changes unless you have more feedback. If you could provide one last review it'd be much appreciated.

Hoping to merge tomorrow after going over the `plugin-release-management` and `sfdx` PRs with @mshanemc`.

.circleci/config.yml
```diff
@@ -15,9 +15,8 @@ jobs:
       - checkout
       - restore_cache: &restore_cache
           keys:
-            - v5-{{checksum ".circleci/config.yml"}}-{{ checksum "yarn.lock"}}
-            - v5-{{checksum ".circleci/config.yml"}}
-            - v5
+            - cache-{{checksum ".circleci/config.yml"}}-{{ checksum "yarn.lock"}}
```
peternhale, line 18: @RodEsp why are the versions being removed?

peternhale (changes requested):

.circleci/config.yml
RodEsp: Because since they're hard-coded they don't actually match up with any particular version of the `oclif` package or the tooling used in the CI build and therefore aren't relevant.

But maybe I'm missing something? Why should we leave them in?

peternhale: Maybe no longer needed, resolving.

peternhale (approved):

PR #711, opened 2021-11-11
feat: support darwin-arm tarballs

### LINKED ISSUE #706: Apple Silicone Support (M1/arm64)
Apple Silicone Support (M1/arm64)
---------------------------------------

[As of version 16, NodeJS supports Apple's new M1 chips](https://doesitarm.com/app/nodejs/). This means that Oclif CLIs should be able to run on these new machines as long as they're using Node v16+.

[Oclif's `pack` commands](https://github.com/oclif/oclif/tree/main/src/commands/pack) don't yet fully support bundling CLI's with the new [NodeJS executables](https://nodejs.org/dist/v16.0.0/). We plan to add this in the foreseeable future but might need help testing it. If someone wants to go ahead and create a PR to get a faster resolution it would be much appreciated!

### Relevant Code:
* https://github.com/oclif/oclif/blob/main/src/tarballs/node.ts#L16-L68
* https://github.com/oclif/oclif/blob/master/src/tarballs/node.ts#L16-L68

mdonnalley:
Adds support for `darwin-arm64` tarballs

Closes #672
Fixes #706 
@W-10115430@


src/commands/pack/tarballs.ts
```diff
@@ -24,6 +24,9 @@ This can be used to create oclif CLIs that use the system node or that come prel
     const {flags} = await this.parse(PackTarballs)
     const targets = flags.targets.split(',')
     const buildConfig = await Tarballs.buildConfig(flags.root, {xz: flags.xz, targets: targets})
+    if (targets.includes('darwin-arm') && !buildConfig.nodeVersion.startsWith('16')) {
```
RodEsp, line 27: node@17 is out now too so we should figure out how to do >16 instead of `startsWith('16')` here.

RodEsp (changes requested):

src/tarballs/config.ts
```diff
@@ -12,6 +12,7 @@ export const TARGETS = [
   'win32-x64',
   'win32-x86',
   'darwin-x64',
+  'darwin-arm'
```
RodEsp, line 15: I'm not sure that just adding this will work because if you look at https://nodejs.org/dist/v16.0.0/ the darwin arm executables are actually named `darwin-arm64` so this will either need to be changed or we might want to have some sort of conversion from different possible names people could use to `darwin-arm64` in the pack command (`darwin-m1`, `apple-silicone`, `m1`, etc....). Whatever you prefer.

mdonnalley: @RodEsp `arm` is translated to `armv7l` here: https://github.com/oclif/oclif/blob/master/src/tarballs/node.ts#L17

So `linux-arm` and `darwin-arm` are actually `linux-armv7l` and `darwin-armv7l`

We could add more explicit targets, e.g.
- `darwin-arm64`
- `darwin-armv7l`
- `linux-arm64`
- `linux-armv7l`

Either way we'd have to keep `linux-arm` for backwards compatibility 

I don't have a strong preference either way

RodEsp: The `armv7l` architecture is for linux 32 bit though, not for darwin 64 bit. We need to differentiate between the two.
There is no such thing as `darwin-armv7l` afaik. Node only has `darwin-arm64` or `darwin-x64` distributables.

image

RodEsp (commented):
Might want to consider putting these changes into `master` too unless we want to force people to upgrade to oclif@v2 to get support for the new Macs.

RodEsp (approved):

peternhale (approved):

src/commands/pack/tarballs.ts
```diff
@@ -25,8 +24,8 @@ This can be used to create oclif CLIs that use the system node or that come prel
     const {flags} = await this.parse(PackTarballs)
     const targets = flags.targets.split(',')
     const buildConfig = await Tarballs.buildConfig(flags.root, {xz: flags.xz, targets: targets})
-    if (targets.includes('darwin-arm64') && !semver.gt(buildConfig.nodeVersion, '16.0.0')) {
-      throw new Error('darwin-arm64 is only supported for node >=16.0.0')
```
RodEsp: This shouldn't be removed. If the user explicitly passes in `darwin-arm64` as a target (i.e. runs `oclif pack:tarballs --targets darwin-arm64`) but is trying to use node 14 or less this should still error. It's only in the case where they don't pass in any targets and they're using node 14 or less that this should just output a warning.

Should just put it above the `buildConfig` assignment instead.

RodEsp (changes requested):

src/commands/pack/tarballs.ts
peternhale: The targets flag defaults to all targets See https://github.com/oclif/oclif/pull/711/files#diff-8c957c644d989965c3dad37783dff849e57acb2933bf0e3f7c9bbb7680866996L16

RodEsp: Ah, good catch. Let's just get rid of the default on the flag then since [`Tarballs.buildConfig()` already uses the defaults if nothing is passed in anyways](https://github.com/oclif/oclif/pull/711/files#diff-8a37a245e4dc9c5cd9977d42aec01025fa1a848c690db79a8dc60087ef607d64R70).

mdonnalley: @RodEsp I'm not sure I understand why we should get rid of the default? At the very least it allow us to communicate in the flag description that the default behavior is to do all targets

RodEsp: @mdonnalley hmm... yes... you're right about conveying the default behavior via the flag help.

My apologies, my suggestions were a little precipitous but overall what I want to see is an error if the user explicitly passes in `darwin-arm64` as a target and is trying to use node 14 or less and just a warning about `darwin-arm64` if the user doesn't pass in any targets and is using node 14 or less, as Pete said.

peternhale: Another alternative would be to alter the default list based on node version. Then on node versions < 16 specifying arm would result in an invalid option error. This way we get errors thrown and target alignment by node version.

peternhale: @mdonnalley @RodEsp have we come to a consensus here so I might continuing with QA?

mdonnalley: @RodEsp Does this work for you? 
Screen Shot 2021-11-16 at 1 17 03 PM

@peternhale help is cached at release time, so we're not able to alter the default based on the node version customers are using

RodEsp (approved):

src/commands/pack/tarballs.ts
RodEsp: @mdonnalley yep that looks fine but will that also happen if the user runs `oclif pack:tarballs --targets darwin-arm64,darwin-x64`?

I'd prefer an error in that scenario as well, anytime the user explicitly passes in `darwin-arm64`.

peternhale:
```
~/oclif-git/oclif/bin/run pack:tarballs --targets darwin-arm64,darwin-x64
 ›   Warning: darwin-arm64 is only supported for node >=16.0.0. Skipping...
```

mdonnalley: @RodEsp so you expect the behavior to be:
- warn and skip `darwin-arm64` if the user does not supply `--targets` flag and node is not 16+
- throw error if `darwin-arm64` is specifically supplied by the user and node is not 16+

What happens if the user supplies all the targets? Do we error or warn?

Personally, I think the existing approach is simpler and more consistent:
- warn and skip `darwin-arm64` if node is not 16+
- throw error if no valid targets are specified

peternhale: There is no easy way to determine if the user supplies `--targets` or it is using the defaults. May not be worth the effort. I support skipping arm with warning.

mdonnalley: Pete's right that there's no way to distinguish between the user supplying all the default targets and the user not supplying the flag at all

At this point, I think we should consider this good enough and wait for feedback from users

RodEsp (approved):

PR #713, opened 2021-11-15
fix: remove dep on lodash.template

peternhale:
@W-10121580@


test/run.js
```diff
@@ -24,7 +24,7 @@ function build(type, features) {
   if (features === 'everything') options = '--options=yarn,typescript,eslint,mocha'
   if (features === 'typescript') options = '--options=yarn,typescript'
   if (features === 'mocha') options = '--options=yarn,mocha'
-  let dir = CI ? tmp.tmpNameSync() : path.join(__dirname, '../tmp')
```
RodEsp, line 27: This seems like a change in functionality, not just a lint fix.
Was this on purpose?

RodEsp (approved):

test/run.js
peternhale: Running tests locally with the original code was causing build errors caused by ts searching up the folder tree and finding the tsconfig for the plugin while trying to run tests for the same project. The local tmp folder is a convenience, not necessary, so always using os tmp dir for tests

PR #723, opened 2021-11-30
fix: bump deps

peternhale:


shetzel (approved):

RodEsp (approved):
I also just changed the branch protection settings to only require the new aggregated test jobs.

PR #764, opened 2021-12-16
feat: optionally include tarball path as input

sumwatshade:
On my team, we have a use case where we are trying to build Oclif within a Bazel environment, where the previous stage has already packaged an NPM tarball. In this case, there is no need for us to rebuild the tarball during the Oclif pack command, and the attempt to do so actually results in weird NPM resolution issues. This feature proposes fixes that issue that provides a flag to input a file path, for example:

```sh
oclif-dev pack --input bazel-bin/packages/my-cli/my_cli.tgz
```


src/commands/pack/deb.ts
```diff
@@ -51,14 +51,15 @@ export default class PackDeb extends Command {
 
   static flags = {
     root: Flags.string({char: 'r', description: 'path to oclif CLI root', default: '.', required: true}),
+    input: Flags.string({ char: 'i', description: 'optionally specify a path to a tarball already generated by NPM', required: false })
```
mdonnalley, line 54: Could you rename this flag to something more descriptive? perhaps `tarball`?

mdonnalley (changes requested):

src/commands/pack/deb.ts
sumwatshade: @mdonnalley done, thank you for review!

mdonnalley (dismissed):

mdonnalley:
@sumwatshade these changes look good to me - but it looks like you're out of date with `main` and have failing tests

sumwatshade:
@mdonnalley I know you merged to main but I have rebased. Is it possible that the failures in the `node-devcli` test is related to this building off of a fork? I am unfamiliar with what this `.circleci/loadgpg` is doing, but I would guess that the environment secret doesn't get transferred to forked PRs for security?

mdonnalley:
@sumwatshade I merged your PR into a separate branch, confirmed the tests were passing, and then merged that into `main`

So your changes should be available in 2.2.0

Thanks for the contribution!

sumwatshade:
Thank you for helping get this PR in :)

PR #766, opened 2021-12-20
doc: fixed minimal node version

pozil:
Doc was reporting node v8 as the minimal version but this is v12 as of v2.0.0.


RodEsp (approved):

RodEsp:
Hey @pozil, thanks for the contribution! We just merged a PR that had some more changes related to the docs that also contained this so will be closing this one. Still much appreciated though!

PR #783, opened 2022-01-05
Restore hook and command generators

### LINKED ISSUE #767: `command is not a oclif command` When trying to create a new command
Possible bug

What is the current behavior?
-----------------------------
Run `oclif generate mynewcli`
`cd mynewcli`
Run `oclif command setup` to create a `setup` command

oclif/2.0.1 darwin-arm64 node-v14.18.0
Big Sur 11.6
When running `oclif command setup` I expect a new `setup` command to be scaffolded

mdonnalley:
- Restores the `hook` and `command` commands from v1 as `generate:hook` and `generate:command`, respectively. 
- Updates REAMDE with notes about migrating from v1 to v2

Fixes #767 

@W-10386078@

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ If you have been using version 1 of the [`oclif` CLI](https://github.com/oclif/o
 
 ## New Commands
 
-Version 2 now includes all the commands in the [`oclif-dev` CLI](https://github.com/oclif/dev-cli) into `oclif`. This means that you can now use a single CLI for all your oclif needs. These commands include:
+Version 2 now includes all the commands from the [`oclif-dev` CLI](https://github.com/oclif/dev-cli). This means that you can now use a single CLI for all your oclif needs. These commands include:
 - `oclif manifest`
 - `oclif pack`
 - `oclif pack:deb`
--- a/README.md
+++ b/README.md
@@ -5,10 +5,8 @@
 oclif: Node.JS Open CLI Framework
 =================================
 
-[![Join the community on Spectrum](https://withspectrum.github.io/badge/badge.svg)](https://spectrum.chat/oclif)
 [![Version](https://img.shields.io/npm/v/oclif.svg)](https://npmjs.org/package/oclif)
 [![CircleCI](https://circleci.com/gh/oclif/oclif/tree/main.svg?style=shield)](https://circleci.com/gh/oclif/oclif/tree/main)
-[![Appveyor CI](https://ci.appveyor.com/api/projects/status/github/oclif/oclif?branch=main&svg=true)](https://ci.appveyor.com/project/heroku/oclif/branch/main)
 [![Downloads/week](https://img.shields.io/npm/dw/@oclif/command.svg)](https://npmjs.org/package/@oclif/core)
 [![License](https://img.shields.io/npm/l/oclif.svg)](https://github.com/oclif/oclif/blob/main/package.json)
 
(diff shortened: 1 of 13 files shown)


RodEsp (approved):

RodEsp (changes requested):
QA
---
    ✅ `oclif generate:command mycommand`
      ✅ Creates a working command
      ✅ Creates a working test
      ✅ `--force` flag overwrites files
      ❌ `--defaults` flag doesn't do anything differently, not needed?
    ✅ `oclif generate:hook myhook` creates a working hook
      ✅ Creates a working hook
      ✅ Creates a working test
      ✅ `--force` flag overwrites files
      ✅ `--event customevent` flag correctly creates a hook under a custom event
      ❌  `--defaults` flag doesn't do anything differently, not needed?
      ❌ This always asks if you want to overwrite the `package.json`. 
          I assume it's because the file exists 
          but it would be a nicer user experience 
          if it only asked you that if the hook already existed and would be overwritten.

    Not sure if this is something we should fix but thought I'd throw it in here to get your thoughts.
    ❌ Running `oclif generate command` tries to generate a CLI called "command" instead of a command
    ❌ Same with `oclif generate hook`

StevenLangbroek:
@RodEsp command generation, a primary functionality, is completely missing. It would appear restoring the preexisting functionality through a semver patch would take precedence over, and be considered separately from ergonomics improvements _to_ those commands, no? In your suggestions appear to be:

- a breaking change: `oclif generate command` should not result in a cli called `command`
- a bugfix: `--defaults` appears to be a noop, and can be deprecated
- an ergonomics improvement: `generate:hook` unnecessarily asks to overwrite package.json

I'd gladly contribute the changes you requested separately if @mdonnalley agrees and/or does not have time, energy or desire to do so.

Would love to hear your thoughts. Thanks 🙇

RodEsp:
@StevenLangbroek, I do agree that restoring these commands is more important than command ergonomics but removing the `--defaults` flag should be so trivial we might as well get it done in this PR.

As for `oclif generate command` and `oclif generate hook` originally I thought it's not such a big deal but since you accurately pointed out that it'd be a breaking change I'd like to avoid having to release Oclif v3 immediately after we just released Oclif v2 if we can help it. With that in mind I'd like to at least explore if it's possible to fix this in this PR to avoid an unnecessary extra major version bump.

As far as having the time and energy to do so, both Mike and I are part of the Salesforce CLI team who actively maintains Oclif as part of our full time job. This PR is part of our current sprint cycle so no worries there. Always happy to accept contributions though so feel free to build on this existing branch if you'd like to speed us up! 🙏🏼

RodEsp (approved):

RodEsp (approved):
QA #2
---
Everything from QA above still works and
```
✅ `oclif generate command mycommand`
  ✅ Creates a working command
  ✅ Creates a working test
  ✅ `--force` flag overwrites files
✅ `oclif generate hook myhook` creates a working hook
  ✅ Creates a working hook
  ✅ Creates a working test
  ✅ `--force` flag overwrites files
  ✅ `--event customevent` flag correctly creates a hook under a custom event
```

StevenLangbroek:
@RodEsp awesome :) question, those QA blocks, are those generated by SF internal tooling?

StevenLangbroek:
@RodEsp got it, looks very cool, nice and concise!

PR #791, opened 2022-01-11
fix: remove fixpack dependency

peternhale:


RodEsp (dismissed):

RodEsp (approved):

RodEsp (changes requested):
QA
---
 ❌ Ran `oclif generate testCLI` and package.json was created with an unexpected and uncommon order.
    🔄 Suggestion: [Lines 82-87](https://github.com/oclif/oclif/pull/791/files#diff-bf715a9a4f63adf855cca52cc2b37014d190db747585b36a9ad333485ee015f6L82-L87) should be reordered so that the required fields, `name` and `version`, appear at the top and the rest generally align to the order shown in [this example](https://docs.npmjs.com/creating-a-package-json-file#example) or in [this guide](https://nodejs.dev/learn/the-package-json-guide).

RodEsp (approved):
QA - round 2
---
✅ Running `oclif generate testCLI` now creates a package.json with the required fields, `name` and `version`, at the top and the rest in a more common order.

PR #792, opened 2022-01-11
fix: correct package name

peternhale:
@W-10270065@


mdonnalley (approved):

RodEsp (approved):
QA
---
✅ Running `oclif generate my-cli` suggests `my-cli` by default as the name for your the new CLI.
✅ It also sets the `name` property in the new `package.json` to that.

PR #458, opened 2020-12-07
chore: upload & promote builds to s3 with new key structure

RasPhilCo:
This PR:
- removes all `publish` commands
- creates the `upload` command
- `upload` uses the CLI version + git sha to push build artifacts to s3 at `versions/<version>/<sha>/<artifact>`
- creates the `promote` command
- `promote` takes channel, version & sha flags and copies the corresponding build manifests/builds to `channel/<channel>/<artifact>` for the `update` command to then pick up

PR #799, opened 2022-01-16
Fix tarballs based commands not considering config from package.json 

rannn505:
- Remove flags defaults from those commands 
- Let `Tarballs.buildConfig` to determine final config
- Fix duplicate `-t` char for `targets` and `tarball` flags in `oclif pack tarballs`


rannn505:
@jdxcode not sure why this check fails 🤔
image

anyway without this fix I have to add --no-xz and --targets to `pack`, `upload`, `promote` commands

rannn505:
@mdonnalley @pimterry @peternhale please help 🙏🏽

src/commands/pack/tarballs.ts
```diff
@@ -11,20 +11,16 @@ This can be used to create oclif CLIs that use the system node or that come prel
 
   static flags = {
     root: Flags.string({char: 'r', description: 'path to oclif CLI root', default: '.', required: true}),
-    targets: Flags.string({char: 't',
-      description: 'comma-separated targets to pack (e.g.: linux-arm,win32-x64)',
-      default: Tarballs.TARGETS.join(','),
-    }),
-    xz: Flags.boolean({description: 'also build xz', allowNo: true, default: true}),
-    tarball: Flags.string({ char: 't', description: 'optionally specify a path to a tarball already generated by NPM', required: false })
+    targets: Flags.string({char: 't', description: 'comma-separated targets to pack (e.g.: linux-arm,win32-x64)'}),
```
mdonnalley, line 14: I would prefer to not remove the `default` here since that would be a breaking change

It might be a better solution to change [this line](https://github.com/oclif/oclif/blob/main/src/tarballs/config.ts#L54) to prefer `updateConfig.node.targets` over `options.targets`. However, that's not a great solution either since user's would never be able to override what's the configured targets

@peternhale @RodEsp Do you have any opinions on this? It looks like the `default` for the `targets` flag was added soon after the `oclif-dev` commands were migrated into `oclif`: https://github.com/oclif/oclif/pull/458

mdonnalley (changes requested):

src/commands/pack/tarballs.ts
rannn505: @mdonnalley Why is this a ״breaking change״?
Correct me if i'm wrong but it seems like if `flags?.targets?` is `undefined` the following line takes `TARGETS` array as a fallback
https://github.com/oclif/oclif/blob/71c3690ea20c333c5cacc2c248510263a80531dd/src/tarballs/config.ts#L70

peternhale: This appears to be the correct; command flag then config then default.

mdonnalley (dismissed):

rannn505:
@mdonnalley I updated this branch from main and resolved all conflicts .. please re-approve this PR and lets handle the failing CI phase so we can go ahead and merge it .. I see more contributors complaining about that 🙏🏽  Thanks.
https://github.com/oclif/oclif/pull/799#issuecomment-1013920930

Regardless of the scope of this PR, I'm willing to get more and more involved and help this project grow if you want me to do so

mdonnalley (approved):

PR #807, opened 2022-01-31
chore: remove github workflows [skip ci]

mdonnalley:


RodEsp:
We shouldn't remove these until unit tests are re-enabled by W-10023776.

mdonnalley:
> We shouldn't remove these until unit tests are re-enabled by W-10023776.

@RodEsp the github workflows aren't doing anything that the circle jobs aren't also doing

https://github.com/oclif/oclif/blob/main/.circleci/config.yml#L86

peternhale (approved):

RodEsp:
@mdonnalley I don't see where CircleCI runs `yarn devcli` which is what the Github workflows are doing.
Am I missing something?

This is the output I would expect to see in the CircleCI job but I can't find that either.
https://github.com/oclif/oclif/runs/5010731216?check_suite_focus=true#step:6:1

PR #808, opened 2022-01-31
chore: enable more tests

peternhale:


package-scripts.js
```diff
@@ -12,13 +12,13 @@ const path = require('path')
 sh.set('-e')
 
 setColors(['dim'])
-
-const testTypes = ['']
+process.env.TEST_SERIES = '1'
+const testTypes = ['', 'other']
```
mdonnalley, line 16: Does `unit` make more sense that `other` here?

peternhale: `other` is really not unit tests - these are the ones that hit aws.

mdonnalley: `integration`?

peternhale: I can do that

mdonnalley (approved):

PR #825, opened 2022-02-09
Fix tarball packing in paths with spaces

pimterry:
This replaces https://github.com/oclif/dev-cli/pull/416.

This PR just adds quotes around a few raw paths used during tarball packing, ensuring that these steps work when run in a path containing spaces.

I've tested this in my project, and everything now seems to build correctly when running `oclif pack tarballs`.


pimterry:
Seems that the build failures are just because `.circleci/loadgpg` can't access a secret (for the `node-devcli` build) and the tests can't access `AWS_ACCESS_KEY_ID` (for the `test-node` builds). Doesn't seem related to this change.

PR #827, opened 2022-02-10
fix: wrongly packaging child bundles on MacOS

IGassmann:
macOS's `pkgbuild` look for valid components in sub-directories when using the `--root` flag. This cause the build process to mark some files in a project to be processed by `pkgbuild` and breaks the resulting CLI.

On my CLI project https://github.com/IGassmann/web-img-optimizer `pkgbuild` would process Chrome which exists in the npm dependency `puppeteer`. Due to that, puppeteer wouldn't work anymore since Chrome was modified by `pkgbuild`.

To avoid that, `pkgbuild` needs to be passed an empty configuration file. It will then not look for valid component in the sub-directories of the project to process them, and will ship the whole directory root as it is.

Here's the [man page of `pkgbuild`](https://www.manpagez.com/man/1/pkgbuild/).


rannn505:
> Looks like there's an issue with CircleCI pipelines.
> 
> CleanShot 2022-02-10 at 10 04 33@2x

[same for me](https://github.com/oclif/oclif/pull/799#issuecomment-1013920930)

IGassmann:
@mshanemc @mdonnalley I've fixed merge conflicts. You're free to review now :)

mdonnalley:
@IGassmann thanks for the PR

We recently switched from CircleCI to Github Actions. Looks like the actions aren't running for any PRs that were open when we made the switch. If it's not a problem, could you open a new PR so that we can get the tests running on your changes? Thanks!

IGassmann:
@mdonnalley sounds good. Here's the new PR: https://github.com/oclif/oclif/pull/971

PR #848, opened 2022-03-14
fix: Add arch to macos pkg file

magarcia:
I just realised from a previous PR https://github.com/oclif/oclif/pull/846 that the `arch` parameter was ignored when building the output filename for macos. 🤦  Packing for x64 and arm64 was overwriting the first file built.


RodEsp:
Hey @magarcia I added some more changes to make sure this is compatible with the rest of oclif commands over at https://github.com/oclif/oclif/pull/849. Would love your thoughts if you have a little time, thanks!

magarcia:
> Hey @magarcia I added some more changes to make sure this is compatible with the rest of oclif commands over at #849. Would love your thoughts if you have a little time, thanks!

Hey @RodEsp thanks for the quick response for those PR's. I missed the `upload` & `promote` commands (we don't use it TBH 😅) , but looks good. 😄 
I'll close this PR in favour of #849

PR #849, opened 2022-03-14
feat!: Add arch to macos pkg file

RodEsp:
Description
---
This adds the ability to build Mac `.pkg` files for [M1 machines](https://en.wikipedia.org/wiki/Apple_M1).

POTENTIAL BREAKING CHANGE:
---
This changes what the mac pkg is named and creates a second one.
We will have to test and make sure that the Mac download link on https://developer.salesforce.com/tools/sfdxcli# will still work after this change.
image


peternhale (dismissed):

RodEsp:
Btw @magarcia it looks like it's going to take us a little longer to merge this than expected but rest assured we're not leaving you hanging. It'll be a breaking change for us because it'll change the URL of the `.pkg` in our S3 buckets which will break our download links.

Not sure if we'll release it as a breaking change in Oclif or not yet, do you have any opinions on the matter?

magarcia:
> Btw @magarcia it looks like it's going to take us a little longer to merge this than expected but rest assured we're not leaving you hanging. It'll be a breaking change for us because it'll change the URL of the `.pkg` in our S3 buckets which will break our download links.
> 
> Not sure if we'll release it as a breaking change in Oclif or not yet, do you have any opinions on the matter?

Ok, no worries. For now we're using a patched version of Oclif to do our releases.

We're facing a similar issue as the one you mention. To keep old URLs it working we plan to upload 3 files, while we update the references to URLs.
- `cli-<version>-<hash>-x64.pkg`
- `cli-<version>-<hash>-arm64.pkg`
- `cli-<version>-<hash>.pkg` (this one is copy ofthe `cli-<version>-<hash>-x64.pkg`)

RodEsp:
Hey @magarcia, I'm currently testing some of this on an M1 and I'm having no problem installing the existing `.pkg` files produced by oclif on it.

It's making me doubt whether we need this at all. Were you encountering some issue with M1 machines or why'd you want to add this kind of support?

magarcia:
> Hey @magarcia, I'm currently testing some of this on an M1 and I'm having no problem installing the existing `.pkg` files produced by oclif on it.
> 
> It's making me doubt whether we need this at all. Were you encountering some issue with M1 machines or why'd you want to add this kind of support?

Hey @RodEsp, sorry for the late reply. I didn't have problems installing the existing `.pkg` but if I remember correctly it was using the x64 version (which is not the optimal one).
With this PR we will be able to create `pkg` installers for x64 and arm64 from an M1 machine.

mdonnalley (dismissed):

mdonnalley (approved):

RodEsp:
Hey @magarcia we ultimately decided to release this as a breaking change so you should now see this new functionality under `oclif@v3.0.0`!

Thanks again for the contribution!

magarcia:
Thanks to you @RodEsp

Issue #902, opened 2022-06-20
Hooks are not functioning


jercle:
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------
Bug
What is the current behavior?
-----------------------------
No hooks are firing when running the cli.
I've also tried starting a project from scratch, same issue.
Have tried using the `init` hook (using `oclif generate hook myhook --event=init`) as well as creating a custom hook using as per docs: https://oclif.io/docs/hooks

<img width="408" alt="image" src="https://user-images.githubusercontent.com/10472533/174532364-09bfaa27-521d-4487-9522-03a0cdad6336.png">

What is the expected behavior?
------------------------------
Please mention your oclif and OS version.
oclif versions:
* oclif/3.0.1 darwin-arm64 node-v16.15.1
* oclif/3.0.1 darwin-arm64 node-v18.4.0
* oclif/2.6.3 darwin-arm64 node-v16.15.1
* oclif/2.6.3 darwin-arm64 node-v18.4.0

OSes: 
* macOS Monterey | 12.1
* Windows 10

Example code: https://github.com/jercle/azgo/tree/45c62d8cacd9f85af10422ca1b4b97e842ff2533/azgoc

Edit: removed freshly generated code as I deleted the repo

jercle:
Found how to enable debugging. The following is shown when debugging:

```bash
  azgoc:azgoc:hooks:init /Users/jercle/git/azgo/azgoc/src/hooks/init/auth.ts:1
  azgoc:azgoc:hooks:init import {Hook} from '@oclif/core'
  azgoc:azgoc:hooks:init ^^^^^^
  azgoc:azgoc:hooks:init 
  azgoc:azgoc:hooks:init SyntaxError: Cannot use import statement outside a module
  azgoc:azgoc:hooks:init     at Object.compileFunction (node:vm:353:18)
  azgoc:azgoc:hooks:init     at wrapSafe (node:internal/modules/cjs/loader:1040:15)
  azgoc:azgoc:hooks:init     at Module._compile (node:internal/modules/cjs/loader:1076:27)
  azgoc:azgoc:hooks:init     at Module._extensions..js (node:internal/modules/cjs/loader:1166:10)
  azgoc:azgoc:hooks:init     at Module.load (node:internal/modules/cjs/loader:988:32)
  azgoc:azgoc:hooks:init     at Module._load (node:internal/modules/cjs/loader:834:12)
[...]

PR #904, opened 2022-06-21
fix: Update hooks generator to point to ./dist/hooks instead of ./src/hooks

jercle:
This will resolve https://github.com/oclif/oclif/issues/902

This does not change where the hooks are saved to as that is done on line 33, it only updates the path in package.json for the end user's cli being built


truesteps:
if you fetch upstream in your repo tests will pass :), I also noticed hooks not working, thanks for the fix, let's hope the devs merge this

jercle:
> if you fetch upstream in your repo tests will pass :), I also noticed hooks not working, thanks for the fix, let's hope the devs merge this

Cheers! Didn't even think hey. Have been a bit under the weather

mdonnalley:
@jercle @truesteps Can you describe the issue you're seeing with using the `src` path?

I just tested this with the following and it works as expected

- `npx oclif generate my-cli`
- `cd my-cli`
- `npx ocli generate hook my_hook --event init`
- `bin/dev hello world --from me`

```
~/code/my-cli » bin/dev hello world --from me                                                                                                                                                                                                    
example hook running hello:world
hello world! (./src/commands/hello/world.ts)
```

mdonnalley:
Never mind - I see the issue now. `bin/dev` knows how to work with the `src` path but `bin/run` doesn't

I'll merge this PR. Thanks for the contribution!

PR #916, opened 2022-07-14
fix: readme for single command cli

### LINKED ISSUE #915: `oclif readme` for single command CLIs produces wrong output
Follows-up oclif/core#426

Like this:

## `bindl .`

Downloads and extracts binaries from compressed packages using a config file

```
USAGE
  $ bindl . [--version] [--help] [-c <value>]
FLAGS
  -c, --config=<value>  Path to the config file
  --help                Show CLI help.
  --version             Show CLI version.
DESCRIPTION
  Downloads and extracts binaries from compressed packages using a config file
  The config will be read from any valid config file in the current directory. The configuration file can be defined
  using all the extensions and names accepted by cosmiconfig, such as bindl.config.js
```

_See code: [src/index.ts](https://github.com/felipecrs/bindl/blob/v2.0.0/src/index.ts)_

mdonnalley:
Fixes `readme` output for single command CLIs

Fixes #915

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "bugs": "https://github.com/oclif/oclif/issues",
   "dependencies": {
-    "@oclif/core": "^1.6.4",
+    "@oclif/core": "^1.9.9",
     "@oclif/plugin-help": "^5.1.12",
     "@oclif/plugin-not-found": "^2.3.1",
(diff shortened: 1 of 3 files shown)


mdonnalley:
> @mdonnalley Is there a test to cover the change?

@peternhale Currently there are no unit tests covering this command (actually, there are multiple commands with no test coverage). I'd rather take that effort on in a separate work item

PR #925, opened 2022-07-27
feat: add `--aliases` flag for readme command

shazron:
This is to allow for `--no-aliases` so the README being generated for a plugin does not include aliases, which may clutter the documentation.

Example usage in `package.json`:
```json
{
  "scripts": {
      "prepack": "oclif manifest && oclif readme --no-aliases"
  }
}
```

### DIFF
--- a/src/commands/readme.ts
+++ b/src/commands/readme.ts
@@ -33,8 +33,6 @@ Customize the code URL prefix by setting oclif.repositoryPrefix in package.json.
     aliases: Flags.boolean({description: 'include aliases in the command list', allowNo: true, default: true }),
   }
 
-  static aliases = ['rdm']
-
   private HelpClass!: HelpBaseDerived
 
   async run(): Promise<void> {
--- a/src/commands/readme.ts
+++ b/src/commands/readme.ts
@@ -30,7 +30,10 @@ Customize the code URL prefix by setting oclif.repositoryPrefix in package.json.
   static flags = {
     dir: Flags.string({description: 'output directory for multi docs', default: 'docs', required: true}),
     multi: Flags.boolean({description: 'create a different markdown page for each topic'}),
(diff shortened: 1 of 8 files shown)


src/commands/readme.ts
```diff
@@ -30,8 +30,11 @@ Customize the code URL prefix by setting oclif.repositoryPrefix in package.json.
   static flags = {
     dir: Flags.string({description: 'output directory for multi docs', default: 'docs', required: true}),
     multi: Flags.boolean({description: 'create a different markdown page for each topic'}),
+    aliases: Flags.boolean({description: 'include aliases in the command list', allowNo: true, default: true }),
   }
 
+  static aliases = ['rdm']
```
mdonnalley, line 36: I'd prefer to not put an alias on this command unless there's overwhelming demand for it

mdonnalley (changes requested):
@shazron thanks for the contribution - this is a great idea!

src/commands/readme.ts
shazron: Me neither, but will have to figure out another way to test this then...

mdonnalley: @shazron You can emulate the tests that use the `fixtures`. For example, https://github.com/oclif/oclif/blob/da5a4d202be114fe7f0c00d2737a9faf8f4e19ab/test/unit/readme.test.ts#L63

Basically, you just need to add a CLI with an alias'd command to the `fixtures` directory and use that for your test case

shazron: @mdonnalley I've resolved it to use a custom fixture. However the tests are failing due to `AWS_ACCESS_KEY_ID` not being set. I assumed the CI would set this in its environment, but it's failing there as well. Those tests are not relevant to this PR and I never touched them.

shazron: Filed https://github.com/oclif/oclif/issues/951

PR #987, opened 2022-10-04
fix: .deb packing privileges

Vesther:
Currently unable to run the debian packing as a non-root user indepentend of sudo privileges.
Because the `chown` and `chgrp` are used for the `root` group anyway, this is my suggested fix for now.


Vesther:
Also @salesforce-cla bot, that email is on my GitHub account 🤔

jurajseffer:
I've ended up removing those two lines completely as `sudo` won't work in containers that don't have it installed.

jdxcode:
huh did I write this code? Seems weird that I made it so sudo was required.

In any case, I think @jurajseffer makes a good point—we can't rely on sudo being installed. I wonder if the `--root-owner-group` flag could be used here to make this step not necessary.

lukepolo:
was this reverted? its requiring it again .

PR #1009, opened 2022-10-20
feat: support JIT plugins

mdonnalley:
Adds commands from JIT plugins to generated manifest

Depends on https://github.com/oclif/core/pull/533

@W-12030328@


KaydenGetsDough:
Help me get plugged in

src/commands/manifest.ts
```diff
@@ -9,13 +21,58 @@ export default class Manifest extends Command {
...
     } catch {}
 
     const {args} = await this.parse(Manifest)
     const root = path.resolve(args.path)
+
+    const packageJson = fs.readJSONSync('package.json') as { oclif: { jitPlugins: Record<string, string> } }
+
+    let jitPluginManifests: Interfaces.Manifest[] = []
+
+    if (flags.jit && packageJson.oclif.jitPlugins) {
```
peternhale, line 45: @mdonnalley I know this is creating the manifest, but it seems that the logic to locate and resolve the correct version might be something useful in core. Thoughts?

Issue #347, opened 2020-05-15
Cannot install `.deb` package: "The repository '/apt ./ Release' does not have a Release file."


busticated:
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------

bug

What is the current behavior?
-----------------------------

1. using an otherwise working oclif cli codebase...
2. produce a `deb` package via `oclif-dev pack:deb`
3. publish your `deb` package via `oclif-dev publish:deb`
4. attempt to install using the instructions provided here: https://oclif.io/docs/releasing#ubuntudebian-packages - specifically:

```
$ wget -qO- https://mys3bucket.s3.amazonaws.com/apt/release.key | apt-key add - # you will need to upload this file manually
$ sudo echo "deb https://mys3bucket.s3.amazonaws.com/apt ./" > /etc/apt/sources.list.d/mycli.list
$ sudo apt update
$ sudo apt install -y mycli
```

...where `mys3bucket` and `mycli` is replaced w/ the appropriate values

What is the expected behavior?
------------------------------

i'd expect the `deb` package to be installed but instead i get:

```
Ign:8 https://o1010000.s3.amazonaws.com/apt ./ InRelease                                                                   
Err:9 https://o1010000.s3.amazonaws.com/apt ./ Release                                                                 
  404  Not Found [IP: 52.217.13.12 443]
...
Reading package lists... Done                                               
E: The repository 'https://o1010000.s3.amazonaws.com/apt ./ Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

using:
`@oclif/dev-cli/1.22.2 linux-x64 node-v12.16.2`
`Ubuntu 18.04.4 LTS`

when i compare to heroku's `deb` setup ([1](https://devcenter.heroku.com/articles/heroku-cli#ubuntu-debian-apt-get), [2](https://cli-assets.heroku.com/install-ubuntu.sh), [3](https://cli-assets.heroku.com/apt/Release), [4](https://cli-assets.heroku.com/apt/Packages.gz)), i appear to be matching every setting. if fact, if you replace `o1010000.s3.amazonaws.com` with `cli-assets.heroku.com` in any of these operations and compare output, you'll see they are the essentially the same. 

when i attempt to debug `apt` using `apt -oDebug::Acquire::https=1 update`, i see logs like:

```
[...]

PR #1035, opened 2022-11-22
feat: remove qq and parallelize build/pack/promote

mshanemc:
we had to drop the --parallel flag on builds because of a bug.

remove `qqjs` and its `qq.cd` that's causing the problem with parallelization.

make sure each build process has its own directory isolation

parallelize all the things

@W-12107459@

### side effects
fix https://github.com/oclif/oclif/issues/347
test: restores integration testing for Debian builds (comment said they weren't runnable on CircleCI. Works on gha)
test: runs integration tests in parallel on different OSes to test both mac and deb packing

---
perf metrics [mac, 2.4 GHz 8-Core Intel Core i9]
1. clear all the cache files `rm -rf dist test/tmp tmp `
2. yarn test:integration

| test        | this PR (ms)           | main (ms) |
| ------------- |:-------------:| :-----:|
| macos | 86,942 | 109,495 |
| tarballs | 281,566  |  696,853  |
| win | 97,470 |  136,077   |

on gha builds (2 core machine).  [mac tests can't run because they only run on macs]

| test        | [this PR ](https://github.com/oclif/oclif/actions/runs/3527800462/jobs/5917250522)(ms)           | [main](https://github.com/oclif/oclif/actions/runs/3506189602/jobs/5873058319#step:9:1) (ms) |
| ------------- |:-------------:| :-----:|
| tarballs | 357,739  |  594,614  |
| win | 67,872 |  101,885   |


src/commands/pack/tarballs.ts
```diff
@@ -18,7 +17,6 @@ This can be used to create oclif CLIs that use the system node or that come prel
   }
 
   async run(): Promise<void> {
-    const prevCwd = qq.cwd()
```
peternhale, line 21: @mshanemc why is managing cwd no longer necessary?

src/tarballs/config.ts
```diff
@@ -32,13 +35,14 @@ export interface BuildConfig {
 
 export async function gitSha(cwd: string, options: {short?: boolean} = {}): Promise<string> {
   const args = options.short ? ['rev-parse', '--short', 'HEAD'] : ['rev-parse', 'HEAD']
-  return qq.x.stdout('git', args, {cwd})
+  const {stdout} = await exec(`git ${args.join(' ')}`, {cwd})
+  return stdout.trim()
```
peternhale, line 39:
```suggestion
  return (await exec(`git ${args.join(' ')}`, {cwd})).stdout.trim();
```

peternhale (dismissed):

src/commands/pack/tarballs.ts
mshanemc: all the paths are now either absolute or relative to the project.  Use of process.exec instead of qq passes in a cwd.

package.json
```diff
@@ -1,145 +1 @@
...
...clif","topicSeparator":" ","macos":{"identifier":"com.oclif.cli"},"update":{"autoupdate":{"rollout":50,"debounce":60},"node":{"version":"16.13.2"},"s3":{"bucket":"dfc-data-production","indexVersionLimit":20,"folder":"media/salesforce-cli/oclif-testing","acl":" ","host":"https://developer.salesforce.com","xz":true}},"topics":{"pack":{"description":"package an oclif CLI into installable artifacts"},"upload":{"description":"upload installable CLI artifacts to AWS S3"}}},"repository":"oclif/oclif","scripts":{"build":"shx rm -rf lib && tsc","devcli:lint":"eslint . --ext .ts --config .eslintrc","devcli:test":"mocha --forbid-only \"test/unit/*.test.ts\"","devcli":"yarn build --noEmit && yarn run devcli:test && yarn run devcli:lint","lint":"nps lint","postpack":"shx rm .oclif.manifest.json","posttest":"yarn run lint","prepack":"shx rm -rf lib && tsc && bin/dev manifest .","test":"nps test && yarn test:unit && yarn test:integration","test:integration":"mocha --forbid-only \"test/integration/*.test.ts\"","test:unit":"mocha --forbid-only \"test/unit/*.test.ts\"","version":"bin/dev readme && git add README.md"},"publishConfig":{"registry":"https://registry.npmjs.org"},"types":"lib/index.d.ts"}
```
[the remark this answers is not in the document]
mshanemc: ahh, sorry.  The integration tests do that.

src/commands/pack/tarballs.ts
mshanemc: nothing ever changes the cwd, which is what was breaking the previous attempt at parallelization

RodEsp (changes requested):
`.DS_Store` shouldn't be checked in. Please remove and add to `.gitignore`.

src/commands/promote.ts
```diff
@@ -105,102 +102,112 @@ export default class Promote extends Command {
...
     }
+
+    await Promise.all(buildConfig.targets.flatMap(target => [
+      promoteManifest(target),
+      promoteGzAndMaybeIndexes(target),
+      maybePromoteXzAndMaybeIndexes(target),
+      maybePromoteMacos(),
+      maybePromoteWin(),
+      maybePromoteDeb(),
+    ]),
```
RodEsp, line 209: This `Promise.all` combined with a `.flatMap` and functions that _maybe_ do something is very complex and unintuitive.  Can you add some comments to clarify what exacly is going on here? 

It also seems like the functions are being called as part of the `.flatMap`? Shouldn't the functions just be returned so that `Promise.all` can run them? And why are the promote macos, win, and deb functions included in this if they don't depend on a `target` being passed into them? Seems like we're just running them an unnecessary amount of times.

I suggest changing the mac, win, and deb functions to not be "maybe" functions and including them like so:
```suggestion
    const promotionTargets = buildConfig.targets.flatMap(target => [
      promoteManifest(target),
      promoteGzAndMaybeIndexes(target),
      maybePromoteXzAndMaybeIndexes(target)
    })
    if (flags.macos) promotionTargets.push(promoteMacos)
    if (flags.win) promotionTargets.push(promoteWin)
    if (flags.deb) promotionTargets.push(promoteDeb)

    await Promise.all(promotionTargets)
```

src/commands/upload/deb.ts
```diff
@@ -33,20 +33,37 @@ export default class UploadDeb extends Command {
       return aws.s3.uploadFile(dist(file), {...S3Options, CacheControl: 'max-age=86400', Key: cloudKey})
     }
 
+    // apt expects ../apt/dists/versionName/[artifacts] but oclif wants varsions/sha/apt/[artifacts]
```
RodEsp, line 36:
```suggestion
    // apt expects ../apt/dists/versionName/[artifacts] but oclif wants versions/sha/apt/[artifacts]
```

.gitignore
```diff
@@ -4,6 +4,10 @@
 /coverage
 /coverage.lcov
 /lib
+/dist
 /node_modules
 /tmp
+/test/tmp
 package-lock.json
+/dist
```
RodEsp, line 12: `/dist` included twice?

RodEsp (changes requested):
Did you do any QA on the promote command? I know integration tests don't exist for it currently but might be nice to add some with all the changes going in.

src/util.ts
```diff
@@ -69,3 +73,21 @@ export const sortVersionsObjectByKeysDesc = (input: VersionsObject): VersionsObj
 
   return result
 }
+
+const homeRegexp = new RegExp(`\\B${os.homedir().replace('/', '\\/')}`, 'g')
+const curRegexp = new RegExp(`\\B${process.cwd()}`, 'g')
```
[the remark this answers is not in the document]
mshanemc: this code came from qq for `prettifyPaths.`. https://github.com/jdxcode/qqjs/blob/5a3357840b7ae730a29f57b0fa9f218ceb26c611/src/log.ts#LL13C16-L13C16

I chose not to mess with it (there weren't comments in the original).

src/commands/promote.ts
mshanemc: good feedback on hitting the installers more than once.  I also took the maybe out of the xz for consistency...see if you like that better.

mshanemc:
> Did you do any testing on the promote command? I know integration tests don't exist for it currently but might be nice to add some with all the changes going in.

just manually.   I checked it again with the new changes via.
```sh
../oclif-repos/oclif-oclif/bin/dev promote --channel qqqa --version 7.179.0 --sha 4383be1
```

I'm looking into the integration test options

src/util.ts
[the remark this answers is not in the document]
RodEsp: 🤔 
I don't think those are actually doing anything... I might test without them and see what happens.

src/commands/promote.ts
```diff
@@ -105,102 +102,109 @@ export default class Promote extends Command {
...
+      const debArtifacts = [
+        templateShortKey('deb', {bin: config.bin, versionShaRevision: debVersion(buildConfig), arch: 'amd64' as any}),
+        templateShortKey('deb', {bin: config.bin, versionShaRevision: debVersion(buildConfig), arch: 'i386' as any}),
+        'Packages.gz',
+        'Packages.xz',
+        'Packages.bz2',
+        'Release',
+        'InRelease',
+        'Release.gpg',
+      ]
```
RodEsp, line 165: Is there any reason to have this outside the `if (flags.dev)` statement?
Would be nice to put it inside.

src/commands/promote.ts
```diff
@@ -105,102 +102,109 @@ export default class Promote extends Command {
...
+    await Promise.all(buildConfig.targets.flatMap(target => [
+      // always promote the manifest and gz
+      promoteManifest(target),
+      promoteGzTarballs(target),
+    ])
+    // optionally promote other artifacts depending on the specified flags
+    .concat(flags.xz ? buildConfig.targets.map(target => promoteXzTarballs(target)) : [])
+    .concat(flags.macos ? [promoteMacInstallers()] : [])
+    .concat(flags.win ? [promoteWindowsInstallers()] : [])
+    .concat(flags.deb ? [promoteDebianAptPackages()] : []),
```
RodEsp, line 207: Nice 👍🏼

RodEsp (dismissed):
Looks good! Just one small comment but doesn't need to be changed to merge.

package.json
```diff
@@ -1,145 +1 @@
...
```
RodEsp, line 1: ???
How'd this lose its formatting? 🤔

test/integration/publish.test.ts
```diff
@@ -1,84 +1,93 @@
...
-        await qq.rm('oclif')
-      }
-
-      await test(manifest.gz, manifest.sha256gz, nodeVersion)
-      await test(manifest.xz, manifest.sha256xz, nodeVersion)
-    }
-
+  .command(['pack:tarballs', '--parallel', '--xz'])
+  .command(['upload:tarballs', '--xz'])
+  .command(['promote', '--channel', pjson.version, '--sha', sha, '--version', pjson.version])
```
RodEsp, line 87: I wouldn't want to create a new channel for every version just to keep things orderly in the s3 bucket, let's just use a `test` channel.

```suggestion
  .command(['promote', '--channel', 'test', '--sha', sha, '--version', pjson.version])
```

test/integration/publish.test.ts
```diff
@@ -1,84 +1,93 @@
...
+  .it('checks uploads for version and channel', async () => {
     await manifest(`versions/${pjson.version}/${sha}`, pjson.oclif.update.node.version)
-    // await testPromote('stable')
+    await manifest(`channels/${pjson.version}`, pjson.oclif.update.node.version)
```
RodEsp, line 90: To go with the previous comment:
```suggestion
    await manifest(`channels/test`, pjson.oclif.update.node.version)
```

mshanemc: no, the tests need to stay isolated, otherwise you'll have flappiness because of collisions on a single channel

package.json
mshanemc: the tests do that (they also modify the version), and then put it back when they're done.  I probably Ctrl-C'd something.

test/integration/publish.test.ts
mshanemc: also, they get cleaned up by the before/after code

src/commands/promote.ts
mshanemc: all the conditional-on-flag logic moved to the Promise.all like your previous suggestion.
https://github.com/oclif/oclif/pull/1035/files#diff-ecee38a5be3a5fc38b4b9db99fbb0b069663f806030445987ec22835d59a034dR205

RodEsp (approved):

RodEsp (approved):

RodEsp (approved):

src/commands/upload/deb.ts
```diff
@@ -33,20 +33,37 @@ export default class UploadDeb extends Command {
       return aws.s3.uploadFile(dist(file), {...S3Options, CacheControl: 'max-age=86400', Key: cloudKey})
     }
 
+    // apt expects ../apt/dists/versionName/[artifacts] but oclif wants versions/sha/apt/[artifacts]
+    // see https://github.com/oclif/oclif/issues/347 for the AWS-redirect that solves this
+    // this workaround puts the code in both places that the redirect was doing
+    // with this, the docs are correct. The copies are all done in parallel so it shouldn't be too costly.
+    const uploadWorkaround = (file: string) => {
+      const cloudKey = `${cloudKeyBase}/apt/./${file}`
```
lukepolo, line 41: Removing the `./` fixes the issue here 

https://github.com/oclif/oclif/issues/1074

PR #1051, opened 2022-12-05
fix: json output formatting

mshanemc:
spaces: 2 on all fse.writeJson

### DIFF
--- a/src/generators/cli.ts
+++ b/src/generators/cli.ts
@@ -218,7 +218,7 @@ export default class CLI extends Generator {
 
     if (_.isEmpty(this.pjson.oclif)) delete this.pjson.oclif
     this.pjson.files = _.uniq((this.pjson.files || []).sort())
-    this.fs.writeJSON(this.destinationPath('./package.json'), this.pjson)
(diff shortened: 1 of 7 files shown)


RodEsp (dismissed):

src/generators/cli.ts
```diff
@@ -1,5 +1,5 @@
 import {execSync} from 'child_process'
-import * as fs from 'fs'
+import * as fs from 'fs-extra'
```
RodEsp, line 2: Should we not change the `writeJSON` on [line 221](https://github.com/oclif/oclif/pull/1051/files#diff-bf715a9a4f63adf855cca52cc2b37014d190db747585b36a9ad333485ee015f6R221) to have the `{spaces: 2}` option passed in?

RodEsp: Or should we leave this import statement alone?

mshanemc: no, "this.fs" is not the same fs.  In fact, I'm gonna switch this import back even though it shouldn't matter

RodEsp (approved):

PR #1092, opened 2023-02-01
fix: workaround wireit

mdonnalley:
wireit does not play nicely with child_process and shelljs. We can work around this by using `yarn tsc` instead of `yarn build` for compiling the JIT plugins during manifest generation.

### DIFF
--- a/src/commands/manifest.ts
+++ b/src/commands/manifest.ts
@@ -61,7 +61,7 @@ export default class Manifest extends Command {
         this.cloneRepo(repoUrl, fullPath, maxSatisfying)
 
         this.executeCommand('yarn', {cwd: fullPath})
-        this.executeCommand('yarn build', {cwd: fullPath})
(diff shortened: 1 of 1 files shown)


src/commands/manifest.ts
```diff
@@ -61,7 +61,7 @@ export default class Manifest extends Command {
         this.cloneRepo(repoUrl, fullPath, maxSatisfying)
 
         this.executeCommand('yarn', {cwd: fullPath})
-        this.executeCommand('yarn build', {cwd: fullPath})
+        this.executeCommand('yarn tsc', {cwd: fullPath})
```
mshanemc, line 64: Can you explain why this didn't work with `yarn build` ?  Is it related to the use of --incremental, or something else?

https://github.com/salesforcecli/plugin-custom-metadata/blob/0434b94bf4c2a3a1a0bd4728b4bc6d99e2b28ecb/package.json#L155

mshanemc (approved):

Issue #1095, opened 2023-02-04
oclif generate command doesn't use the new `Args`


julianduque:
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------

This is definitely a **bug**

What is the current behavior?
-----------------------------
When using `oclif generate command` isn't using the new `Args` class, hence the command fails.

What is the expected behavior?
------------------------------
Template should use the `Args` class

PR #1096, opened 2023-02-04
fix: add Args to command template

julianduque:
This addresses #1095


julianduque:
Thanks @mdonnalley for fixing. I tried to sign the CLA but it failed for some reason (not sure if conflict because I'm employee, not sure)

PR #1110, opened 2023-02-17
feat: Adds an option to specify compression used for pack:deb

colindean:
This enables passing options to override what dpkg-deb uses by default, which is zstd on newer versions of Debian/Ubuntu and xz on older versions. Debian 11 can't read zstd so when building on newer Debian base for older, specifying may be necessary.

This should aid in the resolution of heroku/cli#2240.

N.b. this is ~untested. There's no unit tests for pack:deb, I don't have an AWS account setup to be able to run integration tests, and this is the first Typescript I've written in like 5-6 years.


colindean:
CLA signed.

colindean:
What action can I take to move this forward?

mdonnalley:
@colindean Thanks for the PR and I'm sorry I missed it - I was out on parental leave when this PR was submitted and failed to see it when I was catching up on everything.

I'm going to merge this to a separate feature branch, make some very minor tweaks, and then merge to main from there.

Thanks again for the contribution!

PR #1134, opened 2023-04-04
bin aliases for deb/win

WillieRuemmele:
add bin aliases for `pack:deb`, `pack:win`, `pack:tar`
add tests to try and verify 

@W-12545429@
@W-12545484@
@W-12545491@


src/tarballs/bin.ts
```diff
@@ -2,15 +2,15 @@
 import {Interfaces} from '@oclif/core'
 import * as fs from 'fs'
 import * as path from 'path'
+import {exec} from 'shelljs'
```
mshanemc, line 5: suggestion: native node:exec (promisified) is faster than shelljs

src/tarballs/bin.ts
```diff
@@ -86,5 +80,10 @@ fi
 `, {mode: 0o755})
   }
 
-  await Promise.all([writeWin32(), writeUnix()])
+  await Promise.all([writeWin32(config.bin), writeUnix()])
+  if (process.platform === 'win32') {
```
mshanemc, line 84: could you concat these other writes onto the array that's going to Promise.all (they'd need to be promisified native exec instead of shelljs) ?

mshanemc (approved):

mshanemc:
QA Plan:

For each of the following, check both cli bins
- install on mac via tarball
- install on mac via installer
- install on windows via tarball
- install on windows via installer
- install via tarball in docker container

will not test test debian

Issue #1074, opened 2023-01-06
Bug: upload deb SignatureDoesNotMatch


lukepolo:
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------
Bug

What is the current behavior?
-----------------------------
If the current behavior is a bug, please provide the steps to reproduce.

Env Varaibels Set

```
      AWS_S3_ENDPOINT: ${{ secrets.AWS_S3_ENDPOINT }}
      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
      AWS_S3_FORCE_PATH_STYLE: true
```

1. `oclif pack deb`
 
2. `oclif upload deb`

```
yarn run v1.22.19
$ oclif upload deb
oclif: starting upload of deb artifacts for v1.0.2-6268c76
oclif: s3:uploadFile ./dist/deb/codepier_1.0.2.6268c76-1_amd64.deb s3://cli/versions/1.0.2/6268c76/apt/codepier_1.0.2.6268c76-1_amd64.deb
oclif: s3:uploadFile ./dist/deb/codepier_1.0.2.6268c76-1_amd64.deb s3://cli/versions/1.0.2/6268c76/apt/./codepier_1.0.2.6268c76-1_amd64.deb
oclif: s3:uploadFile ./dist/deb/Packages.gz s3://cli/versions/1.0.2/6268c76/apt/Packages.gz
oclif: s3:uploadFile ./dist/deb/Packages.xz s3://cli/versions/1.0.2/6268c76/apt/Packages.xz
oclif: s3:uploadFile ./dist/deb/Packages.bz2 s3://cli/versions/1.0.2/6268c76/apt/Packages.bz2
oclif: s3:uploadFile ./dist/deb/Release s3://cli/versions/1.0.2/6268c76/apt/Release
oclif: s3:uploadFile ./dist/deb/Packages.gz s3://cli/versions/1.0.2/6268c76/apt/./Packages.gz
oclif: s3:uploadFile ./dist/deb/Packages.xz s3://cli/versions/1.0.2/6268c76/apt/./Packages.xz
oclif: s3:uploadFile ./dist/deb/Packages.bz2 s3://cli/versions/1.0.2/6268c76/apt/./Packages.bz2
[...]

Issue #770, opened 2021-12-28
oclif promote shows NoSuchKey: The specified key does not exist.


tlkiong:
I am not sure if this is a bug. However, based on the docs [here](https://oclif.io/docs/releasing#standalone-tarballs), I should be able to push to S3 by using the command `oclif promote`.

However, after some fiddling around and trying to understand what is going wrong due to lack of documentation, I encounter this issue:

```
    NoSuchKey: The specified key does not exist.
    Code: NoSuchKey
```

RodEsp:
Hey @tlkiong, if you're using Oclif v2 you'll need to use `oclif upload` first to upload your CLI to S3 and then you can use `oclif promote` to promote the CLI to a specific release channel.

We are currently working on updating and improving the docs. Sorry for the hassle.
Let me know if that helps!

creatzor:
I get this error because it attempts to copy `<app-name>-v<version>-<sha>-<target>-buildmanifest` however, the upload command never uploaded that file.

mattbishop:
Same failure for me at the `buildmanifest` file. I would expect `oclif pack tarballs` to generate that file, but it does not.

I am following the directions here: https://oclif.io/docs/releasing

mattbishop:
It looks like oclif expects the `*-buildmanifest` file to be in the S3 bucket already?

https://github.com/oclif/oclif/blob/db03bd0ddc1fedb5930bb21dd4aa75cff035e420/test/integration/publish.test.ts#L52

If so, how do I create this file? What are its contents? Is there an oclif command to generate and push? Feels like there is a missing step in the instructions for tarball releases.

mattbishop:
Just realized this was a test file, but still the question holds--how does one create a `{arch}-buildmanifest` file?

DaanSchoukens:
Wondering the same thing

knvpk:
Facing same problem. is this addressed ?

RodEsp:
Hey @mattbishop, @DaanSchoukens, @creatzor. Looking into this right now.

I think we're missing a check in the `upload` and `promote` commands that exists for `pack`.
Looks like if there's no `s3.host` in the oclif config the `-buildmanifest` files won't be built. See [here](https://github.com/oclif/oclif/blob/main/src/tarballs/build.ts#L120).

Do you have `oclif.update.s3.host` set in your `package.json` as described in the [Autoupdater docs](https://oclif.io/docs/releasing#autoupdater)?

[...]

PR #1140, opened 2023-04-14
fix(promote:deb): skip i386 deb pkg if node < 10

cristiand391:
See: https://github.com/oclif/oclif/issues/770 and https://github.com/oclif/oclif/issues/1074

@W-13037040@

### DIFF
--- a/src/commands/promote.ts
+++ b/src/commands/promote.ts
@@ -168,7 +168,7 @@ export default class Promote extends Command {
       // https://github.com/oclif/oclif/issues/770#issuecomment-1508719530
       const arches = buildConfig.targets.filter(t => t.platform === 'linux')
       
-      if (arches.find(a=> a.arch.includes('x86')) && lte(buildConfig.nodeVersion, 'v9.11.2')) {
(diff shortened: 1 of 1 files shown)


src/commands/promote.ts
```diff
@@ -153,14 +154,25 @@ export default class Promote extends Command {
...
         'Release.gpg',
       ]
+      
+      // start
+      // TODO: remove in next major release
+      // node dropped 32-bit support for linux a long time ago, see:
+      // https://github.com/oclif/oclif/issues/770#issuecomment-1508719530
+      const arches = buildConfig.targets.filter(t => t.platform === 'linux')
+      
+      if (arches.find(a=> a.arch.includes('x86')) && lte(buildConfig.nodeVersion, 'v9.11.2')) {
```
cristiand391, line 171: 9.11.2 is the last known node version that shipped 32bit builds for linux:
https://nodejs.org/dist/latest-v9.x/

I was leaning just remove it from the debArtifacts push but maybe we could break someone?

mshanemc (approved):

PR #1148, opened 2023-05-17
feat: Add support for Linux arm64

magarcia:
This change adds support for arm64 Linux. Although, in theory, the current `arm` (`armel`) architecture should work thanks to [Multiarch](https://wiki.debian.org/Multiarch) (I'm not familiar, so I may not understood that part), it fails in some situations.

For example, `armel` `deb` installers fail on `ubuntu:focal` Docker images that when executed on a Macbook M1 with the following error:
```
dpkg -i cli.deb
dpkg: error processing archive cli.deb (--install):
 package architecture (armel) does not match system (arm64)
Errors were encountered while processing:
 cli.deb
```

The PR also adds support to upload and promote more architectures for Linux (currently was only `amd64` and `i386`)

### DIFF
--- a/src/commands/promote.ts
+++ b/src/commands/promote.ts
@@ -3,11 +3,11 @@ import * as path from 'path'
 import * as _ from 'lodash'
 
 import {ux, Command, Flags} from '@oclif/core'
-import { lte } from 'semver'
+import {lte} from 'semver'
 
 import aws from '../aws'
 import * as Tarballs from '../tarballs'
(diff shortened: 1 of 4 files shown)


mshanemc:
Looks good, thanks for the PR (and good code).  I made one TS tweak to avoid repeating that union type.

moved to a non-fork branch for running tests, your commits are now on https://github.com/oclif/oclif/pull/1158

PR #1155, opened 2023-07-20
Sm/jit-manifest-scripts

mshanemc:
https://github.com/salesforcecli/cli/actions/runs/5606149385

try to keep devScripts from updating pjson files during install.

### explanation 

oclif is building JIT manifests by cloning a repo, yarn install, yarn build, and then oclif-loading a plugin to get its manifest.
https://github.com/oclif/oclif/blob/15c43daf208755f27ed5aee3ad884d4c3fe8f725/src/commands/manifest.ts#L63

sometimes dev-scripts requires yarn install to run twice because it's updating deps.

we shouldn't see this on user-land JIT installs (we don't do all the dev deps, so no dev-scripts to interfere with it).

this JIT manifest process needs devDeps because it's building from the source repo.

### proof that it solves the problem
https://github.com/salesforcecli/cli/actions/runs/5612387454/job/15206000418?pr=995

### side effects

🧹  bumps a bunch of deps, including oclif/test which still had beta versions of oclif/core in it.


iowillhoit:
> including oclif/test which still had beta versions of oclif/core in it.

[Doesn't look like this is true](https://github.com/oclif/test/blob/2.3.31/package.json), did you fix this?

iowillhoit (approved):

PR #1163, opened 2023-07-31
fix: remove yosay dependency

mdonnalley:
- Removes `PackageJson` type in favor of @oclif/core's `Interfaces.PJSON`
- Removes `yosay` dependency
- Removes `oclif2` bin alias used for testing

@W-12153783@
@W-13825265@

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "bugs": "https://github.com/oclif/oclif/issues",
   "dependencies": {
-    "@oclif/core": "^2.9.4",
+    "@oclif/core": "^2.11.4",
     "@oclif/plugin-help": "^5.2.14",
     "@oclif/plugin-not-found": "^2.3.32",
     "@oclif/plugin-warn-if-update-available": "^2.0.44",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -4,12 +4,11 @@
   "version": "3.11.0",
   "author": "Salesforce",
   "bin": {
-    "oclif": "bin/run",
```
mshanemc, line 7: these were kinda handy for testing installers.  Any reason they need to be removed?

package.json
```diff
@@ -87,9 +84,6 @@
       "@oclif/plugin-not-found"
     ],
     "bin": "oclif",
-    "binAliases": [
```
mshanemc, line 90: same comment as above

src/types/index.ts
```diff
@@ -1,16 +1,5 @@
 import * as Generator from 'yeoman-generator'
 
-export interface PackageJson {
```
mshanemc, line 3: nice

package.json
mdonnalley: @mshanemc I didn't like the fact that we were exposing an additional bin for no reason other than tests so I modified the tests to be able to work without needing to have oclif2 bin defined here.

mshanemc (approved):

mshanemc:
QA notes: 

✅  generated a command and a project (plugin/cli with cjs)

PR #1164, opened 2023-08-02
Changing hello example(deprecated) to hello world example

Abhi5Pande:
The hello command does not work by itself. I believe It now requires an accompanying argument, either specifying a person's name or using "world".


mdonnalley (approved):

Abhi5Pande:
Thank you

Issue #1128, opened 2023-03-24
how to remove "See code" from readme?


linojon:
hi, is there an easy way to prevent ```oclif readme```from generating the lines with "See code:..." ?

cristiand391:
Hi @linojon 👋🏼 

`oclif readme` will generate that section only if you have the usage/commands/toc tags.

```
oclif readme --help
adds commands to README.md in current directory

USAGE
  $ oclif readme --dir <value> [--multi] [--aliases]

FLAGS
  --[no-]aliases  include aliases in the command list
  --dir=<value>   (required) [default: docs] output directory for multi docs
  --multi         create a different markdown page for each topic

DESCRIPTION
  adds commands to README.md in current directory
  The readme must have any of the following tags inside of it for it to be replaced or else it will do nothing:
  # Usage
  <!-- usage -->
  # Commands
  <!-- commands -->
  # Table of contents
  <!-- toc -->

  Customize the code URL prefix by setting oclif.repositoryPrefix in package.json.
```

linojon:
Yes thanks I realize that. Im including these tags and want the Usage, Description etc sections. But as a customer facing documentation i do not want the last line the gives a link into my code.

linojon:
Eg

"See code: dist/commands/build.ts"

cristiand391:
Ah, I see. I don't think there's a way to disable just that part but just looked at one of my plugins and I'm not getting that in my README:
https://github.com/cristiand391/sf-plugin-api

I'll check back on Monday 👍🏼

adamu:
Setting `repository` in `package.json` to a full URL, rather than (an presumed GitHub) `owner/repo` string seems to remove the code links - and corresponds with @cristiand391's example. I'm not sure if this is by design though.

cristiand391:
I was looking into this yesterday and found the readme doesn't include the "See code" section in my plugin because of a bug in how oclif is getting the command id, here it splits it by `:`:
https://github.com/oclif/oclif/blob/main/src/commands/readme.ts#L224

but I set spaces as a topic sep here:
https://github.com/cristiand391/sf-plugin-api/blob/main/package.json#L71.

@adamu I'm not sure if that's correct, but after looking at the code in oclif it seems there are a few cases where you can end up in this scenario.

cristiand391:
Hey all, sorry for the late reply.

I opened a PR to fix `oclif readme` missing the `See code` block in plugins using spaces as the topic separator.

[...]

PR #1165, opened 2023-08-02
fix: respect topic sep when generating readme

cristiand391:
Closes https://github.com/oclif/oclif/issues/1128

`oclif readme` would miss the `See code` block in CLIs using spaces as a topic separator.

Test:
1) clone https://github.com/cristiand391/sf-plugin-api
2) run `yarn oclif readme`, no `See code` in readme
3) yarn link this PR into the plugin, repeat step 2.

@W-13872809@

### DIFF
--- a/src/commands/readme.ts
+++ b/src/commands/readme.ts
@@ -221,7 +221,7 @@ USAGE
   private commandPath(plugin: Interfaces.Plugin, c: Command.Cached): string | undefined {
     const commandsDir = plugin.pjson.oclif.commands
     if (!commandsDir) return
-    let p = path.join(plugin.root, commandsDir, ...c.id.split(':'))
+    let p = path.join(plugin.root, commandsDir, ...c.id.split(plugin.pjson.oclif.topicSeparator ?? ':'))
     const libRegex = new RegExp('^lib' + (path.sep === '\\' ? '\\\\' : path.sep))
     if (fs.pathExistsSync(path.join(p, 'index.js'))) {
       p = path.join(p, 'index.js')
--- a/src/commands/readme.ts
+++ b/src/commands/readme.ts
@@ -173,11 +173,14 @@ USAGE
     }
 
     try {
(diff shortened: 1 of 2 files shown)


src/commands/readme.ts
```diff
@@ -221,7 +221,7 @@ USAGE
   private commandPath(plugin: Interfaces.Plugin, c: Command.Cached): string | undefined {
     const commandsDir = plugin.pjson.oclif.commands
     if (!commandsDir) return
-    let p = path.join(plugin.root, commandsDir, ...c.id.split(':'))
+    let p = path.join(plugin.root, commandsDir, ...c.id.split(plugin.pjson.oclif.topicSeparator ?? ':'))
```
cristiand391, line 224: `plugin.pjson.oclif.topicSeparator` instead of `this.config.topicSeparator` because we want the sep in the CLI, not the one from `oclif`. Default to `:` if not defined.

mdonnalley: The `id` property should _always_ be separated by colons, meaning you shouldn't have to lookup the topicSeparator here 

It looks like the actual issue here is that the [`formatCommand`](https://github.com/oclif/core/blob/61aa44e5ae5e200d5df69433fdf1b8bd6cd9f2dc/src/help/index.ts#L242) method in oclif/core is mutating the command id to respect the configured `topicSeparator`.

You could consider updating `formatCommand` to make a copy of the command before altering the `id`, although I don't know if that would have downstream consequences in `Help`. A simpler solution might be to make a copy of the command before passing it to `formatCommand` on line 179

mdonnalley (changes requested):

src/commands/readme.ts
cristiand391: Ahh, good catch. I made it copy the command here instead of in oclif/core 👍🏼.

mdonnalley (approved):

PR #1187, opened 2023-09-20
feat: improve perf of manifest command

mdonnalley:
For jit plugins `manifest` now downloads and extracts the npm package to get the `oclif.manifest.json` for the specified version. This is significantly faster than cloning and building each JIT plugin's repository

@W-14122700@

### DIFF
--- a/src/commands/manifest.ts
+++ b/src/commands/manifest.ts
@@ -116,6 +116,12 @@ export default class Manifest extends Command {
 
   private getVersion(plugin: string, version: string): string {
     if (version.startsWith('^') || version.startsWith('~')) {
+      // Grab latest from npm to get all the versions so we can find the max satisfying version.
+      // We explicitly ask for latest since this command is typically run inside of `npm prepack`,
+      // which sets the npm_config_tag env var, which is used as the default anytime a tag isn't
+      // provided to `npm view`. This can be problematic if you're building the `nightly` version
+      // of a CLI and all the JIT plugins don't have a `nightly` tag themselves.
+      // TL;DR - always ask for latest to avoid potentially requesting a non-existent tag.
       const {versions} = JSON.parse(this.executeCommand(`npm view ${plugin}@latest --json`).stdout) as {
         versions: string[]
       }
--- a/src/commands/manifest.ts
+++ b/src/commands/manifest.ts
@@ -4,6 +4,12 @@ import * as path from 'path'
 import * as os from 'os'
 import * as semver from 'semver'
 import {exec, ShellString, ExecOptions} from 'shelljs'
+import got from 'got'
+import {promisify} from 'util'
+import {pipeline as pipelineSync} from 'stream'
+import {checkFor7Zip} from '../util'
+
+const pipeline = promisify(pipelineSync)
 
 async function fileExists(filePath: string): Promise<boolean> {
   try {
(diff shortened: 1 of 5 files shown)


src/commands/manifest.ts
```diff
@@ -47,25 +53,33 @@ export default class Manifest extends Command {
...
 
-        const versions = JSON.parse(this.executeCommand(`npm view ${jitPlugin}@latest versions --json`).stdout)
-        const maxSatisfying = semver.maxSatisfying(versions, version)
+        await fs.mkdir(fullPath, {recursive: true})
 
-        this.cloneRepo(repoUrl, fullPath, maxSatisfying)
+        const tarballUrl = this.getTarballUrl(jitPlugin, version)
+        const tarball = path.join(fullPath, path.basename(tarballUrl))
+        await pipeline(
+          got.stream(tarballUrl),
```
mshanemc, line 66: if anyone else in the oclifiverse ever starts doing JIT stuff, and is building behind a proxy or via CI, they'll have to ask us to enable proxy support for this.

src/commands/manifest.ts
```diff
@@ -99,16 +113,30 @@ export default class Manifest extends Command {
...
-      this.executeCommand(`git clone --branch ${tag} ${repoUrl} ${fullPath} --depth 1`)
-    } catch {
-      try {
-        this.executeCommand(`git clone --branch v${tag} ${repoUrl} ${fullPath} --depth 1`)
-      } catch {
-        throw new Error(`Unable to clone repo ${repoUrl} with tag ${tag}`)
+  private getTarballUrl(plugin: string, version: string): string {
+    // jit plugin is unpinned so we need to figure out the max satisfying version
+    if ((version.startsWith('^') || version.startsWith('~'))) {
+      const npmLatest = JSON.parse(this.executeCommand(`npm view ${plugin}@latest --json`).stdout) as {
```
mshanemc, line 119: `@latest` has no effect on the output and can be removed.  It's also confusing because we don't want latest, we want everything (ex: the jit version is `^4.0.0-beta.0`)

src/commands/manifest.ts
```diff
@@ -99,16 +113,30 @@ export default class Manifest extends Command {
...
+      return dist.tarball
     }
+
+    // jit plugin is pinned so we don't need to figure out the max satisfying version
+    const {dist} = JSON.parse(this.executeCommand(`npm view ${plugin}@${version} --json`).stdout) as {
+      versions: string[]
+      dist: { tarball: string }
+    }
+    return dist.tarball
+
```
mshanemc, line 139: this is a bit repetitive.

suggestion: 
- 1 function to calculate the version (pinned version if pinned, use your new `maxSatisfying` technique otherwise)
- 1 function to get the tarball, given a version

mshanemc (approved):
non-blocking suggestions.

I'm ok not doing proxy until someone needs it.

src/commands/manifest.ts
mdonnalley: `@latest` is in there because of an interesting npm feature we discovered while trying to release the JIT feature. Here's the PR that introduced it: https://github.com/oclif/oclif/pull/1107

`oclif manifest` gets run during `prepack` which means `npm` has set `npm_config_tag` to the requested tag (could be `nightly`, `latest-rc`, or `latest`), which is used for any `npm` command. So if we don't explicitly set `latest` then it will try to request a tag that doesn't exist for our plugins

If you request `latest`, you'll still get all the versions - even the ones that are ahead of latest. For instance, `npm view @oclif/core@latest versions` shows all the v3 beta versions even though latest is currently 2.15.0

src/commands/manifest.ts
mdonnalley: Possibly. Although this exact same technique is used for downloading and extracting the node executable, which hasn't been an issue for anyone yet

src/commands/manifest.ts
```diff
@@ -99,16 +114,30 @@ export default class Manifest extends Command {
...
-      } catch {
-        throw new Error(`Unable to clone repo ${repoUrl} with tag ${tag}`)
+  private getVersion(plugin: string, version: string): string {
+    if (version.startsWith('^') || version.startsWith('~')) {
+      // Grab latest from npm to get all the versions so we can find the max satisfying version.
+      // We explicitly ask for latest since this command is typically run inside of `npm prepack`,
+      // which sets the npm_config_tag env var, which is used as the default anytime a tag isn't
+      // provided to `npm view`. This can be problematic if you're building the `nightly` version
+      // of a CLI and all the JIT plugins don't have a `nightly` tag themselves.
+      // TL;DR - always ask for latest to avoid potentially requesting a non-existent tag.
```
iowillhoit, line 124: lol, I read this whole thing before getting to the TL;DR 😅

iowillhoit:
## QA NOTES

- 🟢 Speeds significantly faster locally 4:11 vs 0:19!
- 🟢 Installed an older version of the CLI and old JIT plugin was used
  - Used `sf@2.2.1`
  - That specifies `plugin-dev@1.1.3`
  - That version was found in my TMPDIR

```
$ cat salesforce-plugin-dev/package/package.json
{
    "name": "@salesforce/plugin-dev",
    "description": "commands for sf plugin development",
    "version": "1.1.3",
    [...]
```

- 🟡 Slight changes on pre/post manifest
  - Default on `target-dev-org`
    - This is probably intentional, but wanted to point it out in case
  - Scanner is quite different
    - This is likely because they are behind on some libraries
  - I'll share some files

Issue #1190, opened 2023-09-27
Bug: `oclif readme` broken


hirasso:
### What is the current behavior?

`oclif readme` doesn't work correctly with typescript projects. It either doesn't output any commands (pre-build) or it outputs incorrect *_See code...* links (post-build).

Let's use the [hello world](https://github.com/oclif/hello-world) template as an example. Installing it locally and then running `oclif readme` will produce incorrect and inconsistent commands docs:

1. When runnning `oclif readme` without having previously built the package, the available commands won't be listed at all (since the script depends on the `oclif.commands` entry in the `package.json` that is (correctly) pointing to the **non-existent dist files** in `/dist/commands`).
2. When first executing `npm run build` (now the `/dist/` folder is present!), `oclif readme` will find the commands, but the links are off, for example: 
> _See code: [dist/commands/hello/index.ts](https://github.com/oclif/hello-world/blob/v0.0.0/dist/commands/hello/index.ts)_

Note the leading `dist/`, but the file extension `.ts`. That file will never exist.

### What is the expected behavior
`oclif readme` should always output the commands with correct source links, pre- and post-build.

### Suggested solution

There needs to be a way to tell oclif where the actual source files are. Maybe something like `oclif.commandsSrc` or an optionally nested object for `oclif.commands` could help:
```json
{
  "oclif": {
    "commands": {
      "dist": "./dist/commands",
      "src": "./src/commands"
    }
  },
}
```

mdonnalley:
@hirasso Thanks for creating this issue. Looks like the `readme` command expects the compiled src directory to be `lib` and will replace `lib` for `src` when generating the command's path. 

I'll create a PR today to make it compatible with `dist`

hirasso:
@mdonnalley did I understand you correctly that `oclif readme` should work correctly if the compiled commands folder is present and called `/lib`? 

Out or curiosity I tested just that: 

- Changed the folder to `/lib`
- ran `npm run build` (compiled code landed in /dist)
- ran `npx oclif readme`

The result looks the same as before to me. One example (for a command called `swup create`):

```
_See code: [lib/commands/create.js](https://github.com/swup/cli/blob/v5.0.1/lib/commands/create.js)_
```

Issues we are having with the above:

- `lib/commands/create.js` could be mangled code, since it's compiled from `/src`
- the version in the URL (`v5.0.1`) is not prepended with a `v` in our package
- `/lib/` should be replaced with `/src/`
- `create.js` should be `create.ts`

[...]

PR #1191, opened 2023-09-27
fix: dont assume lib as outDir

mdonnalley:
- Warn users when running `readme` on plugin with no compiled source
- Check `tsconfig.json` for the `outDir` so that it can be replaced with `src` when generating the command's code path

### DIFF
--- a/src/commands/readme.ts
+++ b/src/commands/readme.ts
@@ -39,11 +39,20 @@ Customize the code URL prefix by setting oclif.repositoryPrefix in package.json.
 
   private HelpClass!: HelpBaseDerived
   private flags!: Interfaces.InferredFlags<typeof Readme.flags>;
+  private outDir!: string;
 
   async run(): Promise<void> {
     this.flags = (await this.parse(Readme)).flags
     const cwd = process.cwd()
     const readmePath = path.resolve(cwd, 'README.md')
+    const tsConfigPath = path.resolve(cwd, 'tsconfig.json')
+    const tsConfig = await fs.readJSON(tsConfigPath).catch(() => ({}))
+    this.outDir = tsConfig.compilerOptions?.outDir ?? 'lib'
+
+    if (!await fs.pathExists(this.outDir)) {
+      this.warn(`No compiled source found at ${this.outDir}. Some commands may be missing.`)
+    }
+
     const config = await Config.load({root: cwd, devPlugins: false, userPlugins: false})
 
     try {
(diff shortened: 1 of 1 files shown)


hirasso:
@mdonnalley in our setup, we are running version updates using a GitHub action, where the compiled source never will be available. What would be the correct way to this then? Add `npm run build` to the action, just before the call to `readme update`? This is the action we are using right now: https://github.com/swup/cli/blob/master/.github/workflows/version-update.yml

iowillhoit:
## QA NOTES

- 🟢 Readme source links render correctly with `src` in a plugin (`plugin-info`)
  - I updated the `tsconfig.json > outDir` to `dist`
  - I updated `package.json` `oclif.commands` to `"./dist/commands"`
  - Ran `yarn build`
  - Ran `yarn oclif readme`
  - Confirmed source links had `dist` replaced with `src`
- 🔴 This breaks links on our top level CLI (which uses `dist`)
  - `yarn link oclif` in your local `cli`
  - Run `yarn oclif readme`
  - Notice that all source links still have `lib`
  - This gets more complicated if plugins have a combination of `lib` and `dist`

I think the easiest solution is pull the first directory off of each plugin's `oclif.commands` and used that in the `libRegex`. Example `"commands": "./dist/commands",` would find `dist` and replace that with `src`. This way each plugin could have it's own `outDir` and a top level CLI would correctly build its source links.

iowillhoit (changes requested):
See QA notes

mdonnalley:
@iowillhoit thanks for the QA. I just pushed a fix for it. 

@hirasso I also pushed a fix for the issues you were describing [here](https://github.com/oclif/oclif/issues/1190#issuecomment-1739018877). The problem was that `readme` was checking for `typescript` in the `devDependencies` to determine if it should replace the compiled dir with the source dir and if it should replace `.js` with `.ts`. Turns out your CLI has `typescript` in the `dependencies` so I just modified the check to look in both places for `typescript`.

src/commands/readme.ts
```diff
@@ -227,26 +234,27 @@ USAGE
   private commandPath(plugin: Interfaces.Plugin, c: Command.Cached): string | undefined {
     const commandsDir = plugin.pjson.oclif.commands
     if (!commandsDir) return
+    const hasTypescript = plugin.pjson.devDependencies?.typescript || plugin.pjson.dependencies?.typescript
     let p = path.join(plugin.root, commandsDir, ...c.id.split(':'))
-    const libRegex = new RegExp('^lib' + (path.sep === '\\' ? '\\\\' : path.sep))
+    const outDir = path.dirname(commandsDir.replace('./', ''))
```
iowillhoit, line 239: Should this use `plugin.root + path.sep` instead of `./`?

iowillhoit (dismissed):
One more comment. QA looks good otherwise!

iowillhoit (approved):

hirasso:
Thanks guys, working great now! 🙏

PR #1194, opened 2023-10-02
Prerelease/v4

mdonnalley:
## Breaking Changes

- Requires Node 18+
- Removes 32-bit node + linux support (See https://github.com/oclif/oclif/issues/770#issuecomment-1508719530)
- Removes `OCLIF_NEXT_VERSION` env var

## Maintenance 
- Removes usage of `shelljs` in production
- Adds commit linting
- Adds prettier (and makes everything pretty)
- Adds husky hooks
- Bumps various linting libraries
- Bumps to typescript v5
- Removes nps for testing
- Adds nyc for unit test code coverage

@W-14222403@


.mocharc.json
```diff
@@ -1,13 +1,16 @@
 {
+  "recursive": true,
+  "reporter": "spec",
   "require": [
     "test/helpers/init.js",
-    "ts-node/register",
-    "source-map-support/register"
```
mshanemc, line 5: we don't need source-map anymore?

bin/dev.js
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env ts-node
```
mshanemc, line 1: why is dev.js `ts-node` but run.js `node`

bin/dev.js
```diff
@@ -0,0 +1,6 @@
...
+// eslint-disable-next-line node/shebang
```
mshanemc, line 2: why?

package.json
```diff
@@ -1,70 +1,71 @@
...
+    "mocha": "^10.2.0",
+    "nyc": "^15.1.0",
+    "prettier": "^3.0.3",
+    "shelljs": "^0.8.5",
     "shx": "^0.3.4",
-    "tmp": "^0.2.1",
     "ts-node": "^10.7.0",
-    "typescript": "4.5.5"
-  },
-  "resolutions": {
```
mshanemc, line 60: good!

package.json
```diff
@@ -1,70 +1,71 @@
...
-    "fancy-test": "^1.4.10",
+    "eslint-config-oclif-typescript": "^2.0.1",
+    "eslint-config-oclif": "^5.0.0",
+    "eslint-config-prettier": "^9.0.0",
+    "eslint": "^8.50.0",
+    "fancy-test": "^3.0.1",
     "globby": "^11.1.0",
-    "mocha": "^9.2.2",
-    "npm-run-path": "^4.0.1",
-    "nps": "^5.10.0",
```
mshanemc, line 54: good!

src/aws.ts
```diff
@@ -31,82 +35,89 @@ const aws = {
   },
   get s3() {
     try {
-      cache.s3 = cache.s3 || new (require('aws-sdk/clients/s3') as typeof S3)({
-        ...this.creds,
-        endpoint: process.env.AWS_S3_ENDPOINT,
-        s3ForcePathStyle: Boolean(process.env.AWS_S3_FORCE_PATH_STYLE),
-      })
+      cache.s3 =
+        cache.s3 ||
```
mshanemc, line 39:
```suggestion
        cache.s3 ??
```

src/commands/pack/macos.ts
```diff
@@ -1,21 +1,20 @@
...
 
 import * as Tarballs from '../../tarballs'
 import {templateShortKey} from '../../upload-util'
-import {exec as execSync} from 'child_process'
-import {promisify} from 'node:util'
-import * as os from 'os'
+import {uniq} from '../../util'
 
 const exec = promisify(execSync)
 type OclifConfig = {
```
mshanemc, line 13: here, and above where you create types for jitPlugins:

Would it be better to update the pjson/oclif typing in `oclif/core` so that `oclif/oclif` can reference those?

src/commands/pack/win.ts
```diff
@@ -228,47 +236,6 @@ the CLI should already exist in a directory named after the CLI that is the root
...
-      const windows = (config.pjson.oclif as any).windows as {name: string; keypath: string; homepage?: string}
-      if (windows && windows.name && windows.keypath) {
-        await signWindows(o, arch, config, windows)
-      } else this.debug('Skipping windows exe signing')
-
-      this.log(`built ${o}`)
-    }))
-  }
-
   private async checkForNSIS() {
```
mshanemc, line 239: I guess it's generally alphabetizing object properties?  It's weird not having `run` as the first method on a command.

And putting that private checkForNSIS first is...weird.

src/commands/readme.ts
```diff
@@ -29,152 +29,179 @@ Customize the code URL prefix by setting oclif.repositoryPrefix in package.json.
...
+        .join('\n')
+        .trim() + '\n'
+    )
   }
 
   renderCommand(config: Interfaces.Config, c: Command.Cached): string {
     this.debug('rendering command', c.id)
-    const title = template({config, command: c})(c.summary || c.description || '').trim().split('\n')[0]
-    const help = new this.HelpClass(config, {stripAnsi: true, maxWidth: columns})
+    const title = template({command: c, config})(c.summary || c.description || '')
```
mshanemc, line 201:
```suggestion
    const title = template({command: c, config})(c.summary ?? c.description ?? '')
```

mshanemc (approved):
approved with minimal comments.  I couldn't read this very well because of the `prettier` noise (but that certainly makes it easier to read in the future!)

ALSO, this probably won't release when you merge it because of the [skip-ci] commits and will need a manual merge

package.json
```diff
@@ -1,70 +1,71 @@
...
+  "version": "4.0.0-beta.1",
   "author": "Salesforce",
   "bin": {
-    "oclif": "bin/run"
+    "oclif": "bin/run.js"
   },
   "bugs": "https://github.com/oclif/oclif/issues",
   "dependencies": {
-    "@oclif/core": "^2.11.4",
+    "@oclif/core": "^3.0.0-beta.22",
```
mshanemc, line 11: 3 is out now, so this can move from beta to that.

bin/dev.js
mdonnalley: dev.js needs ts-node to be able to transpile the ESM src code at run time. Another option would be to use `node --loader=ts-node/esm`

.mocharc.json
mdonnalley: I don't see a need for it

bin/dev.js
mdonnalley: eslint complains with `this file needs no shebang`, which isn't true

mshanemc (approved):

PR #1210, opened 2023-11-16
fix: build-time and install-time options for windows defender

mshanemc:
1. box is checked by default (non-breaking change)
2. you can specify `checked` | `unchecked` | `hidden` (hidden sets the value to false)

testing files from CLI `https://developer.salesforce.com/media/salesforce-cli/oclif-testing/defender-check/...`

[@W-14495012@](https://gus.lightning.force.com/a07EE00001eb2V9YAI)
https://github.com/forcedotcom/cli/issues/2298


src/commands/pack/win.ts
```diff
@@ -220,6 +227,9 @@ export default class PackWin extends Command {
 the CLI should already exist in a directory named after the CLI that is the root of the tarball produced by "oclif pack:tarballs"`,
       hidden: true,
     }),
+    'hide-defender-option': Flags.boolean({
```
mdonnalley, line 230: there's an `env` flag option that will default to what's in the environment if a value is provided. I think that would allow you to skip the env var check on line 89

mdonnalley (changes requested):
Small change + what we discussed in slack about CLI developers being able to maintain the current functionality if they so choose

src/commands/pack/win.ts
mshanemc: oh, it can used the scopedEnvVar thing?

mdonnalley: hm, yeah good point - it's not a scoped env var

src/commands/pack/win.ts
```diff
@@ -220,6 +234,13 @@ export default class PackWin extends Command {
 the CLI should already exist in a directory named after the CLI that is the root of the tarball produced by "oclif pack:tarballs"`,
       hidden: true,
     }),
+    'defender-exclusion': Flags.string({
```
mdonnalley, line 237:
```typescript
'defender-exclusion': Flags.option({
  options: ['checked', 'unchecked', 'hidden'] as const,
})({
  summary: 'set to...',
  // etc...
}),
```

mdonnalley (approved):

src/commands/pack/win.ts
mshanemc: oh, nice.

mdonnalley (approved):

PR #1229, opened 2024-01-02
fix: missing core yarn files

ericis:
Newer versions of yarn depend on more files than "yarn.lock". As a result, the tarballs need to include these core files to ensure that any future commands against the tarball contents utilize the right version of `yarn`.

## Example

Configuring yarn v4 with a "pinned version" requires the ".yarnrc.yml" file along with the `yarnPath` configuration value. If configured to use a specific version of yarn, Yarn will download required files into a "./.yarn/releases/" directory. In the future, any execution of `yarn` will pick up the `yarnPath` configuration and use the "./.yarn/releases/" directory contents to download and use the specific version of yarn.

## Current impacts

Without these changes, modern versions of `yarn`, like Yarn v4, do not properly install their dependencies when packaging for release. For example, if yarn 4 is configured to scope dependency installations to the workspace and use a classic node modules setup, the current build will only copy "yarn.lock" and either not upgrade Yarn to v4 and not copy (respect) the v4 configuration from ".yarnrc.yml" or will properly discover and use yarn v4 via `corepack` from the "package.json" file and "packageManager" value, but will still not copy (respect) the v4 configuration from ".yarnrc.yml".

## Testing these changes

1. An existing oclif CLI was "monkey-patched" with these changes by editing the generated oclif file "my-cli\node_modules\oclif\lib\tarballs\build.js" with a slightly modified version of these same changes.
2. `yarn build`
3. `yarn oclif pack win`

### DIFF
--- a/src/tarballs/build.ts
+++ b/src/tarballs/build.ts
@@ -1,6 +1,6 @@
 import {Interfaces} from '@oclif/core'
 import * as findYarnWorkspaceRoot from 'find-yarn-workspace-root'
-import {copy, emptyDir, move, readJSON, remove, writeJSON} from 'fs-extra'
+import {copy, emptyDir, mkdir, move, readJSON, remove, writeJSON} from 'fs-extra'
 import {exec as execSync} from 'node:child_process'
 import {existsSync} from 'node:fs'
 import {mkdir, readdir, rm} from 'node:fs/promises'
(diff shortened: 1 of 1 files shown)


ericis:
> Thanks for the contribution! Before we can merge this, we need @ericis to [sign the Salesforce Inc. Contributor License Agreement](https://cla.salesforce.com/sign-cla?prUrl=https%3A%2F%2Fgithub.com%2Foclif%2Foclif%2Fpull%2F1229).

This is now signed.

mdonnalley (approved):

mdonnalley:
@ericis Thanks for this contribution! I'm going to merge this into a feature branch that we can run all our integration tests against and then merge to `main` once the tests pass

Issue #1232, opened 2024-01-03
Relative command path in package.json does not replace outDir with src


ClayChipps:
Summary
-----------------------------

Whenever the `commandPath` function is called within the `readme` command, there is an issue if the user has specified a relative path in their package.json for the `plugin.pjson.oclif.commands`. Specifically, the RegEx pattern does not match, and the outDir is not replaced with `src` as expected.

What is the current behavior?
-----------------------------

As an example, many of the [salesforcecli projects](https://github.com/salesforcecli/plugin-packaging/blob/main/package.json#L49) use this type of "commands" key in the package.json -> `"commands": "./lib/commands",`.

When I use this similar convention there is an issue with the RegEx that is being used [here](https://github.com/oclif/oclif/blob/main/src/commands/readme.ts#L262). Specifically, the command path will start with `./lib` for example, but the [path value](https://github.com/oclif/oclif/blob/main/src/commands/readme.ts#L260) will be `lib/commands/my/command/etc`. Thus, the RegEx does not find a match and it does not replace lib with src.

I was able to "resolve" this issue by changing the value of `commands` from `./lib/commands` to `lib/commands` in the package.json. While this solves the issue, I was moreso curious why so many of the official salesforcecli projects use the `./lib/commands` in their package.json files? I was able to recreate this issue using the plugin-packaging project on my machine (windows).

What is the expected behavior?
------------------------------

I would expect to have both `./lib/commands` and `lib/commands` work equally. If you all agree with this, I would be happy to work on a PR to adjust the behavior. I didn't want to do so until I understood if this is an expected behavior or not.

OS: Windows 11 Pro 22621.2861 
oclif: 4.1.1

mdonnalley:
@ClayChipps Thanks for posting the issue

Looks like this is an issue only on windows. All of our release actions for the Salesforce plugins run on Ubuntu, which is why we haven't noticed this before.

If you're up for it, a PR would be awesome!

ClayChipps:
Hey @mdonnalley I submitted #1043 which should resolve this issue. Appreciate your quick response to the issue!

PR #1234, opened 2024-01-04
fix: readme command does not parse path correctly on windows

ClayChipps:
This fixes issue #1232. I replaced the string pattern in the call to `commandDir.replace` with a RegEx that will accept either `./` or `.\`. As per the [node docs](https://nodejs.org/api/path.html#pathsep), the only two possible values for `path.sep` are / for POSIX and \ for Windows. Thus, I don't see any issue replacing the path.sep with the RegEx to check for the / or \ characters.


mdonnalley (approved):

mdonnalley:
Thanks @ClayChipps for doing this 🏆

PR #1236, opened 2024-01-08
fix: outdir regex in readme

mdonnalley:
Fixes the regex used for finding the `outDir` during readme generation


mshanemc (dismissed):
yep, that fixes it

PR #1238, opened 2024-01-08
feat: add node.options

royra:
Allow specifying [options to the Node process](https://nodejs.org/api/cli.html#options) when creating the CLI binary in the `pack:tarballs` command.

Example of useful options include:
- [`--max-old-space-size=...`](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes) to increase memory size
- [`--disable-warning=...`](https://nodejs.org/api/cli.html#--disable-warningcode-or-type) to disable warnings, e.g, `ExperimentalWarning`

The list of options can be specified as a string array or space-separated string in the `oclif.update.node.options` section of the `package.json`:

example:

```json
{
  "oclif": {
    "update": {
      "node": {
        "options": ["--max-old-space-size=4096", "--no-warnings"]
      }
    }
  }
}
```

or:

```json
{
  "oclif": {
    "update": {
      "node": {
        "options": "--max-old-space-size=4096 --no-warnings"
      }
    }
  }
}
```

### DIFF
--- a/src/tarballs/bin.ts
+++ b/src/tarballs/bin.ts
@@ -10,10 +10,12 @@ const exec = promisify(execSync)
 export async function writeBinScripts({
   baseWorkspace,
   config,
+  nodeOptions,
   nodeVersion,
 }: {
   baseWorkspace: string
   config: Interfaces.Config
+  nodeOptions: string[]
   nodeVersion: string
 }): Promise<void> {
   const binPathEnvVar = config.scopedEnvVarKey('BINPATH')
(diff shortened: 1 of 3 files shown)


mdonnalley:
@royra Great idea! I'd like to merge this in but I'd prefer for the setting to apply to all the `pack` commands first

royra:
> @royra Great idea! I'd like to merge this in but I'd prefer for the setting to apply to all the `pack` commands first

@mdonnalley The other `pack` commands use the user-provided scripts at the `bin` dir, which in turn execute node. The options can be provided there:

`bin/run` or `bin/run.js` for ESM:
```bash
#!/usr/bin/env -S node --max-old-space-size=4096
# body of the script follows
```

`bin/run.cmd`:
```cmd
@echo off
node --max-old-space-size=4096 "%~dp0\run.js" %*
```

mdonnalley:
Apologies for not knowing this part of the code base as well as I should. After some digging, it looks like the other pack commands (`deb`, `macos`, and `win`) all build a tarball (using the same code path as `pack tarballs`) before wrapping it in the os-specific installer. So the change you've made will already apply to all those as well.

mdonnalley (approved):

PR #1215, opened 2023-12-07
feat: upgrade to use aws sdk v3

Amndeep7:
resolves https://github.com/oclif/oclif/issues/1108

PR #1240, opened 2024-01-09
feat: upgrade to use aws api v3 (#1215)

mdonnalley:
Merges #1215


Amndeep7:
@mdonnalley seems like your CI might need to pass AWS_REGION as well in https://github.com/oclif/oclif/blob/main/.github/workflows/test.yml for each of the platforms

mdonnalley:
QA: Tested that this version is able to successfully publish `sf`: https://github.com/salesforcecli/cli/actions/runs/7466357735

PR #1262, opened 2024-01-21
fix: make eslint plugin a dev dependency

G-Rath:
This was mistakenly added as a dependency in #1194, when it should be a dev dependency


G-Rath:
@mdonnalley it would be great if we could get this landed and released since it has a big impact on downstream consumers (especially using `npm` or `pnpm` since they install peer dependencies by default, so this results in ESLint getting installed in production builds), and is prone to conflicts due to the dependabot PRs

mdonnalley (approved):

mdonnalley:
Thanks @G-Rath for finding and fixing this 🏆

PR #1264, opened 2024-01-24
feat: deprecate and hide lock command

mdonnalley:
Deprecate and hide the `lock` command since the `oclif.lock` feature will not be supported in the next major version of `@oclif/plugin-plugins` (see https://github.com/oclif/plugin-plugins/pull/776)

### DIFF
--- a/src/commands/lock.ts
+++ b/src/commands/lock.ts
@@ -24,6 +24,9 @@ export default class Lock extends Command {
     this.hidden = true
     this.state = 'deprecated'
     this.summary = 'Copy the yarn.lock to oclif.lock'
+    this.deprecationOptions = {
+      message: 'oclif.lock is deprecated and will be removed in @oclif/plugin-plugins v5.0.0 and above.',
+    }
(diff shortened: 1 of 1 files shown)


iowillhoit:
QA looks good. Before:
Screenshot 2024-01-26 at 1 51 12 PM

After:
Screenshot 2024-01-26 at 1 51 57 PM


🟡 Setting the `deprecationOptions.message` might help users see when this is going into effect. Something like:
```js
this.deprecationOptions = {
      message: 'oclif.lock is deprecated and will be removed in @oclif/plugin-plugins v5.0.0 and above',
}
```

iowillhoit (approved):

iowillhoit:
Looks good:
```
> bin/run.js lock
 ›   Warning: oclif.lock is deprecated and will be removed in @oclif/plugin-plugins v5.0.0 and above.
Copying yarn.lock to oclif.lock
```

iowillhoit (approved):

PR #1283, opened 2024-02-04
Update win.ts and fix error by adding quotation marks

lukasdatte:
If the path to the project contains a whitespace, makensis fails without any info. Adding quotation marks will fix the failing part.


lukasdatte:
I signed the cla but there has been an error. Would you check that?

mdonnalley (approved):

PR #1322, opened 2024-03-07
fix: force local paths when extracting on win32

otaviojacobi:
Currently oclif/oclif uses a lot of bash in order to build applications and natively does not support windows runners such as cmd and pwsh. However, this can be worked around using wsl or github actions `bash` runner (even on windows machines).

Almost everything works and allows us to build with it in these bash environments, however, the only step that fails is this extract step because a Windows path will look like `C:\some\path` and tar will natevely interpret anything with a `:` as an IP: https://superuser.com/questions/1720172/what-does-tar-cannot-connect-to-resolve-failed-mean

We can't use this for win/macOS runners however as the `tar` cli in there does not support it: https://github.com/r-lib/remotes/issues/171

This just adds the --force-local flag to windows builds making much easy to build on native windows runners.

### DIFF
--- a/src/tarballs/build.ts
+++ b/src/tarballs/build.ts
@@ -91,7 +91,9 @@ export async function build(
     await emptyDir(c.workspace())
     const tarballNewLocation = path.join(c.workspace(), path.basename(tarball))
     await move(tarball, tarballNewLocation)
-    await exec(`tar -xzf "${tarballNewLocation}"`, {cwd: c.workspace()})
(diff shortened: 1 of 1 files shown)


mdonnalley (approved):

mdonnalley:
Thanks for the contribution @otaviojacobi 

I'm going to merge this into a separate feature branch so that all the integration tests can run (they don't run on forks). Once those pass, I'll merge this into main.

Issue #999, opened 2022-10-14
Add oclif to existing project


toymachiner62:
Do you want to request a *feature* or report a *bug*?
-----------------------------------------------------
Request a feature (unless it already exists)

What is the current behavior?
-----------------------------
I can't run `oclif generate myprojec` and point to my existing project and have just the necessary files added there.

What is the expected behavior?
------------------------------
Essentially I want a way to add oclif to an existing project

joshcanhelp:
I can take a crack at this one. From poking around and trying things out, it seems like the two things that need to be done, at a minimum, are:

- Install `@oclif/core` if not
- Add the `./bin/*` files. Because those are different between the [CJS](https://github.com/oclif/hello-world/tree/main/bin) and [ESM](https://github.com/oclif/hello-world-esm/tree/main/bin) examples, we'll need to ask the user what module type they're using and maybe try to [auto-detect](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_enabling).
- Add the following to `package.json`:

```js
{
	// ...
	"oclif": {
		"commands": "./dist/commands"
	}
}
```

Questions:

- Do we need to do anything else in this command? From my testing, this was the bare minimum to get `./bin/run.js command` working.
- Are there any other config values we should be setting in `package.json`?

CC @mdonnalley

mdonnalley:
> we'll need to ask the user what module type they're using and maybe try to [auto-detect](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_enabling).

I think detecting the module type would be the best user experience here

> Are there any other config values we should be setting in package.json?

`oclif generate` will also set the `bin` which gets set in `oclif.bin` and in `bin`. It's not necessary for running `bin/run.js` but will be necessary once it's published. 

If the package.json already has a `bin` you could use that to populate `oclif.bin`

PR #1358, opened 2024-04-06
feat: Add init command

joshcanhelp:
This adds a new command, `init`, that adds the necessary files and configuration to an existing npm project. It attempts to auto-detect the module type and package manager being used, falling back to a prompt if necessary. Usage is covered in the examples. 

Addresses #999

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -117,9 +117,7 @@ export default class Generate extends GeneratorCommand<typeof Generate> {
 
     const repoBinPath = join(repoPath, 'bin')
     const projectBinPath = join(location, 'bin')
-    if (!statSync(projectBinPath).isDirectory()) {
-      await mkdir(projectBinPath)
-    }
+    await mkdir(projectBinPath, {recursive: true})
 
     const binFiles = await readdir(repoBinPath)
     for (const binFile of binFiles) {
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1,8 +1,8 @@
 /* eslint-disable unicorn/no-await-expression-member */
 import {Errors, Flags} from '@oclif/core'
 import chalk from 'chalk'
-import {existsSync, renameSync, statSync} from 'node:fs'
-import {readdir, rm, writeFile} from 'node:fs/promises'
+import {accessSync, renameSync, statSync} from 'node:fs'
+import {constants, mkdir, readdir, rm, writeFile} from 'node:fs/promises'
 import {join, resolve, sep} from 'node:path'
 
 import {FlaggablePrompt, GeneratorCommand, exec, makeFlags, readPJSON} from '../generator'
@@ -30,7 +30,14 @@ const FLAGGABLE_PROMPTS = {
   },
   'output-dir': {
     message: 'Directory to build the CLI in',
-    validate: (d: string) => existsSync(resolve(d)),
+    validate(d: string) {
+      try {
+        accessSync(resolve(d), constants.X_OK)
+        return true
+      } catch {
+        return false
+      }
+    },
   },
   'package-manager': {
     message: 'Select a package manager',
@@ -109,11 +116,16 @@ export default class Generate extends GeneratorCommand<typeof Generate> {
     await clone(template, repoPath)
 
     const repoBinPath = join(repoPath, 'bin')
+    const projectBinPath = join(location, 'bin')
+    if (!statSync(projectBinPath).isDirectory()) {
+      await mkdir(projectBinPath)
+    }
+
     const binFiles = await readdir(repoBinPath)
     for (const binFile of binFiles) {
       const filePath = join(repoBinPath, binFile)
       if (statSync(filePath).isFile()) {
-        renameSync(filePath, join(location, 'bin', binFile))
+        renameSync(filePath, join(projectBinPath, binFile))
       }
     }
 
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1,24 +1,12 @@
 /* eslint-disable unicorn/no-await-expression-member */
 import {Errors, Flags} from '@oclif/core'
 import chalk from 'chalk'
-import {accessSync, renameSync, statSync} from 'node:fs'
-import {constants, mkdir, readdir, rm, writeFile} from 'node:fs/promises'
+import {accessSync} from 'node:fs'
+import {constants, readdir, writeFile} from 'node:fs/promises'
 import {join, resolve, sep} from 'node:path'
 
 import {FlaggablePrompt, GeneratorCommand, exec, makeFlags, readPJSON} from '../generator'
 import {validateBin} from '../util'
-
-async function clone(repo: string, location: string): Promise<void> {
-  try {
-    await exec(`git clone https://github.com/oclif/${repo}.git "${location}" --depth=1`)
-  } catch (error) {
-    const err =
-      error instanceof Error
-        ? new Errors.CLIError(error)
-        : new Errors.CLIError('An error occurred while cloning the template repo')
-    throw err
-  }
-}
 
 const validModuleTypes = ['ESM', 'CommonJS']
 const validPackageManagers = ['npm', 'yarn', 'pnpm']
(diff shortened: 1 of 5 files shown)


src/commands/init.ts
```diff
@@ -0,0 +1,167 @@
...
+    message: 'Select a package manager:',
+    options: ['npm', 'yarn', 'pnpm'],
+    validate: (d: string) => ['npm', 'pnpm', 'yarn'].includes(d) || 'Invalid package manager',
+  },
+} satisfies Record<string, FlaggablePrompt>
+
+export default class Generate extends GeneratorCommand<typeof Generate> {
+  static description = `This will clone the template repo and update package properties. For CommonJS, the 'oclif/hello-world' template will be used and for ESM, the 'oclif/hello-world-esm' template will be used.`
+
+  static examples = [
```
joshcanhelp, line 42: TODO: Update these once we're happy with the flags

src/commands/init.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+const FLAGGABLE_PROMPTS = {
+  bin: {
+    message: 'Command bin name the CLI will export:',
+    validate: (d: string) => validateBin(d) || 'Invalid bin name',
+  },
+  'output-dir': {
+    message: 'Directory to build the CLI in:',
+    validate: (d: string) => existsSync(resolve(d)),
```
joshcanhelp, line 30: Using `existsSync` here to conform to the types (`async` not allowed)

src/commands/init.ts
```diff
@@ -0,0 +1,167 @@
...
+  },
+  'package-manager': {
+    message: 'Select a package manager:',
+    options: ['npm', 'yarn', 'pnpm'],
```
joshcanhelp, line 34: Could DRY this array

src/commands/init.ts
```diff
@@ -0,0 +1,167 @@
...
+      throw new Errors.CLIError(`Could not find a package.json file in ${location}`)
+    }
+
+    const bin = await this.getFlagOrPrompt({
+      defaultValue: location.split(sep).at(-1)!,
+      name: 'bin',
+      type: 'input',
+    })
+
+    let moduleType = this.flags['module-type']
```
joshcanhelp, line 89: This logic allows the flag to be respected as-is, then auto-detection, then ask.

src/commands/init.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (!moduleType) {
+      if (packageJSON.type === 'module') {
+        moduleType = 'ESM'
+      } else if (packageJSON.type === 'commonjs') {
+        moduleType = 'CommonJS'
+      } else {
+        moduleType = await (
```
joshcanhelp, line 96: I used the code from `getFlagOrPrompt` as an example for this.

src/commands/init.ts
```diff
@@ -0,0 +1,167 @@
...
+      description: 'Supply answers for specific prompts and accept default values for the rest',
+    },
+  ]
+
+  static flaggablePrompts = FLAGGABLE_PROMPTS
+
+  static flags = {
+    ...makeFlags(FLAGGABLE_PROMPTS),
+    'module-type': Flags.option({
+      options: ['CommonJS', 'ESM'],
```
joshcanhelp, line 66: Could DRY this array

src/commands/init.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    const template = moduleType === 'ESM' ? 'hello-world-esm' : 'hello-world'
+    const repoPath = join(location, '_tmp-template')
+    await clone(template, repoPath)
+
+    const repoBinPath = join(repoPath, 'bin')
+    const binFiles = await readdir(repoBinPath)
+    for (const binFile of binFiles) {
+      const filePath = join(repoBinPath, binFile)
+      if (statSync(filePath).isFile()) {
```
joshcanhelp, line 118: Using sync here because async fails aslant checks. Options here are:

- Keep this as-is
- Use `async` and ignore or change eslint config
- Explicitly move the 4 files

src/commands/init.ts
```diff
@@ -0,0 +1,163 @@
...
+        dirname: bin,
+      },
+    }
+
+    await writeFile(join(location, 'package.json'), JSON.stringify(updatedPackageJSON, null, 2))
+
+    const installedDeps = Object.keys(packageJSON.dependencies || {})
+    if (!installedDeps.includes('@oclif/core')) {
+      let packageManager = this.flags['package-manager']
+      if (!packageManager || !validPackageManagers.includes(packageManager)) {
```
joshcanhelp, line 140: Same as module type, this allows us to accept the flag, then auto-detect, then ask if we don't have anything to go with.

src/commands/init.ts
```diff
@@ -0,0 +1,164 @@
...
+    await rm(repoPath, {recursive: true})
+
+    const updatedPackageJSON = {
+      ...packageJSON,
+      bin: {[bin]: './bin/run.js'},
+      oclif: {
+        bin,
+        commands: './dist/commands',
+        dirname: bin,
+        ...packageJSON.oclif,
```
joshcanhelp, line 132: I have this at the bottom to override anything that the init command is doing. This seems like the right thing to do in an existing project.

src/commands/init.ts
```diff
@@ -0,0 +1,164 @@
...
+    ...makeFlags(FLAGGABLE_PROMPTS),
+    'module-type': Flags.option({
+      options: validModuleTypes,
+    })({}),
+  }
+
+  static summary = 'Initialize a new oclif CLI'
+
+  async run(): Promise<void> {
+    const outputDir = await this.getFlagOrPrompt({defaultValue: './', name: 'output-dir', type: 'input'})
```
joshcanhelp, line 73: I went back and forth on whether this should be a flag or an arg. I landed on flag to better match how `generate` works but I don't feel strongly either way.

src/commands/init.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    this.log(`Initializing oclif in ${chalk.green(location)}`)
+
+    const packageJSON = (await readPJSON(location))!
+    if (!packageJSON) {
+      throw new Errors.CLIError(`Could not find a package.json file in ${location}`)
+    }
+
+    const bin = await this.getFlagOrPrompt({
+      defaultValue: location.split(sep).at(-1) || '',
```
joshcanhelp, line 84: Tries to get the current directory name as a reasonable default.

src/commands/init.ts
```diff
@@ -0,0 +1,164 @@
...
+      name: 'bin',
+      type: 'input',
+    })
+
+    let moduleType = this.flags['module-type']
+    if (!moduleType || !validModuleTypes.includes(moduleType)) {
+      if (packageJSON.type === 'module') {
```
joshcanhelp, line 91: I think this is about as good as we can do with module auto-detection here. We could parse `tsconfig.json` ([module](https://www.typescriptlang.org/tsconfig#module) seemed to be the strongest signal) but I 'm not sure we want to go that far with something that doesn't require TS.

mdonnalley (changes requested):
Looking good! Just a few questions/changes

I'll checkout this out later today to test it out

src/commands/init.ts
joshcanhelp: I think it's good to allow tools like this to operate in other directories, like `npm --prefix /path/to/project` or `git -C /path/to/project`

src/commands/init.ts
```diff
@@ -0,0 +1,174 @@
...
+          message: 'Select a module type',
+        })
+      }
+    }
+
+    this.log(`Using module type ${chalk.green(moduleType)}`)
+
+    const template = moduleType === 'ESM' ? 'hello-world-esm' : 'hello-world'
+    const repoPath = join(location, '_tmp-template')
+    await clone(template, repoPath)
```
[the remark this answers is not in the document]
joshcanhelp: I think that would be _much_ better than cloning a repo and hoping the permissions and everything works.

Would that make maintenance on your template repos more arduous?

src/commands/init.ts
```diff
@@ -0,0 +1,174 @@
...
+          packageManager = await (
+            await import('@inquirer/select')
+          ).default({
+            choices: validPackageManagers.map((manager) => ({name: manager, value: manager})),
+            message: 'Select a package manager',
+          })
+        }
+      }
+
+      await exec(`${packageManager} install @oclif/core`, {cwd: location, silent: false})
```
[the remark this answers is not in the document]
joshcanhelp: Good catch, I haven't used yarn in a while ... looks like that's the same in `pnpm`

src/commands/init.ts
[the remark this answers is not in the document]
mdonnalley: Yeah that would be the downside but we already have to maintain the other templates to keep up with the templates so it's not a new problem.

I think I'd prefer templates for now. We can always solve that problem later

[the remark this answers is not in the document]
joshcanhelp: @mdonnalley - Just to be clear ... are you saying you want me to use templates added to this repo for this PR?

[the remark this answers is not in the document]
mdonnalley: Yes, I would prefer for you to add templates for all the bin scripts to the `templates/` directory.

You'll just need to insert `__dirname` or `import.meta.url` based on the module type

joshcanhelp:
@mdonnalley - Alright, we should be good here 👍

mdonnalley (approved):
This looks good to me - thanks again for the contribution!

I'm going to merge this into a separate branch before merging to main so I can add some integration tests before it's released.

Depending on what the tests reveal, I might make some modifications but your initial commit will still be in the history so you'll get attributed with the bulk of the work. Hopefully that works for you

PR #1366, opened 2024-04-09
feat: add init command (#1358)

### LINKED ISSUE #999: Add oclif to existing project

mdonnalley:
Merges #1358 

Fixes #999

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1,3 +1,4 @@
+import select from '@inquirer/select'
 import {Errors, Flags} from '@oclif/core'
 import chalk from 'chalk'
 import {readdir, writeFile} from 'node:fs/promises'
@@ -6,10 +7,22 @@ import {join, resolve, sep} from 'node:path'
 import {FlaggablePrompt, GeneratorCommand, exec, makeFlags, readPJSON} from '../generator'
 import {validateBin} from '../util'
 
+const VALID_PACKAGE_MANAGERS = ['npm', 'yarn', 'pnpm'] as const
+type PackageManager = (typeof VALID_PACKAGE_MANAGERS)[number]
+
+function isPackageManager(d: string): d is PackageManager {
+  return VALID_PACKAGE_MANAGERS.includes(d as PackageManager)
+}
+
 const FLAGGABLE_PROMPTS = {
   bin: {
     message: 'Command bin name the CLI will export',
     validate: (d: string) => validateBin(d) || 'Invalid bin name',
+  },
+  'package-manager': {
+    message: 'Select a package manager',
+    options: VALID_PACKAGE_MANAGERS,
+    validate: (d: string) => isPackageManager(d) || 'Invalid package manager',
   },
   'topic-separator': {
     message: 'Select a topic separator',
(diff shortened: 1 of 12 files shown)


mdonnalley:
@joshcanhelp FYI I've made a few changes. Feel free to push back on any of them

- added integration tests to ensure this works across package managers, operating systems, and module types
- made `--output-dir` a normal flag instead of a promptable flag. I think it'll be a better UX for `oclif init` to simply default to the current working directory without prompting
- removed the `--module-type` flag. I think it's safe to work off the `type` in the package.json and I can't imagine a reasonable scenario where someone with a CommonJS project would want to init oclif with the ESM bin scripts (or vice versa)
- added a `--topic-separator` promptable flag. I think it'll be nice for people to have that option set for them, especially since oclif defaults to colons - which few people actually want.
- run `chmod` on the bin scripts so that they're runnable
- install `ts-node` if it's not already there since it's needed for `bin/dev.js`
- a few code preference/quality changes

Again, feel free to push back on any of these changes

src/commands/init.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (!installedDeps.includes('@oclif/core')) {
+      this.log('Installing @oclif/core')
+      await exec(`${packageManager} ${packageManager === 'yarn' ? 'add' : 'install'} @oclif/core`, {
+        cwd: location,
+        silent: false,
+      })
+    }
+
+    const allInstalledDeps = [
+      ...Object.keys(packageJSON.dependencies ?? {}),
```
joshcanhelp, line 174:
```suggestion
      ...installedDeps,
```

src/commands/init.ts
```diff
@@ -0,0 +1,187 @@
...
+  return select({
+    choices: VALID_PACKAGE_MANAGERS.map((manager) => ({name: manager, value: manager})),
+    default: 'npm',
+    message: 'Select a package manager',
+  })
+}
+
+export default class Generate extends GeneratorCommand<typeof Generate> {
+  static description =
+    'This will add the necessary oclif bin files, add oclif config to package.json, and install @oclif/core if needed.'
```
joshcanhelp, line 57:
```suggestion
    'This will add the necessary oclif bin files, add oclif config to package.json, and install @oclif/core and ts-node.'
```

src/commands/init.ts
```diff
@@ -0,0 +1,187 @@
...
+    {
+      command: '<%= config.bin %> <%= command.id %>',
+      description: 'Initialize a new CLI in the current directory.',
+    },
+    {
+      command: '<%= config.bin %> <%= command.id %> --output-dir "/path/to/existing/project"',
+      description: 'Initialize a new CLI in a different directory.',
+    },
+    {
+      command: '<%= config.bin %> <%= command.id %> --topic-separator colons --bin mycli.',
```
joshcanhelp, line 69:
```suggestion
      command: '<%= config.bin %> <%= command.id %> --topic-separator colons --bin mycli',
```

mdonnalley:
>My only hesitation here is that you can use .mjs and .cjs extensions to indicate the module type that you want to use without type being in the package.json ([v18 docs](https://nodejs.org/docs/latest-v18.x/api/esm.html#enabling) and [v20 docs](https://nodejs.org/docs/latest-v20.x/api/esm.html#enabling) and others going back).

That's a good point that I hadn't considered. I'll add it back 👍

PR #1385, opened 2024-04-25
feat: no lockfiles in tarballs by default

mshanemc:
`pack:tarballs` gets a new `--prune-lockfiles` flag.  
- If present, tarballs command will remove tarballs (oclif.lock, yarn.lock, npm-shrinkwrap,  package-lock) from the local project before creating the tarball.

This is done after `addDependencies` so that the lockfiles affect that step.

https://github.com/forcedotcom/cli/discussions/2835


src/commands/pack/tarballs.ts
```diff
@@ -9,6 +9,7 @@ Add a pretarball script to your package.json if you need to run any scripts befo
 
   static flags = {
     parallel: Flags.boolean({description: 'Build tarballs in parallel.'}),
+    'preserve-lockfiles': Flags.boolean({description: 'Preserve lockfiles in the tarball.'}),
```
mdonnalley, line 12: What's the argument for this being an opt-out flag? As opposed to an opt-in flag like `--remove-lockfiles`

src/tarballs/build.ts
```diff
@@ -72,186 +72,194 @@ const copyCoreYarnFiles = async (yarnRootPath: string, workspacePath: string) =>
...
+    for (const target of targetsToBuild) {
+      // eslint-disable-next-line no-await-in-loop
+      await buildTarget(target, c, options)
+    }
+
+    log(`finished building ${targetsToBuild.length} targets sequentially`)
   }
+}
+
+const isLockFile = (f: string) =>
```
mdonnalley, line 114: `pnpm-lock.yaml`?

mshanemc: cool, didn't know about that one, will add.

src/commands/pack/tarballs.ts
mdonnalley: I think my preference is to make it an opt-in flag so that we don't risk suddenly breaking other people's builds - even if it's a small risk.

I have a list of of things to do in the next major. If we want, we can switch to an opt-out flag then

mdonnalley (approved):

PR #1421, opened 2024-05-24
feat: use @oclif/core v4

mdonnalley:
Use @oclif/core v4

@W-15845231@

### DIFF
--- a/src/commands/pack/win.ts
+++ b/src/commands/pack/win.ts
@@ -363,8 +363,12 @@ async function signWindows(
   o: string,
   arch: string,
   config: Interfaces.Config,
-  windows: Interfaces.PJSON.Plugin['windows'],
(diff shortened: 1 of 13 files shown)


src/commands/pack/win.ts
```diff
@@ -363,8 +363,12 @@ async function signWindows(
   o: string,
   arch: string,
   config: Interfaces.Config,
-  windows: Interfaces.PJSON.Plugin['windows'],
+  windows: Interfaces.OclifConfiguration['windows'],
 ) {
+  if (!windows) {
+    throw new Error('windows not set in package.json')
```
iowillhoit, line 369: This would be misleading if they are using an rc file. Maybe say "windows not set in oclif configuration" ??

iowillhoit (approved):

iowillhoit (approved):

Issue #1526, opened 2024-08-29
Can't build tarballs within pnpm repos


randallt21:
**Describe the bug**
In a repo using pnpm, building and packing tarballs using the oclif script fails. Script execution: `oclif pack tarballs`, or for me `pnpm oclif pack tarballs`. 

**To Reproduce**

1. Be in a pnpm repo
2. Run `pnpm oclif pack tarballs`
3. Exits with error 
	`Error: ENOENT: no such file or directory, stat '/Users/randall/git/extend-cli/npm-shrinkwrap.json'`
    `Code: ENOENT`

Code causing the error is here: https://github.com/oclif/oclif/blob/2f54ced402cd88feb8a9ae63457f82da4320fc4d/src/tarballs/build.ts#L173-L178

**Expected behavior**
This should package and build like it would in yarn and npm.

**Your Environment (please complete the following information):**

- `oclif` Version: 4.14.26
- OS & version: MacOS Sonoma
- Shell/terminal & version: zsh 5.9

**Additional context**
I'm happy to setup a repo to repro in if necessary. I have resolved this in my local fork and tested it in our app, successfully packing the tarball locally.
<img width="727" alt="Screenshot 2024-08-29 at 4 41 18 PM" src="https://github.com/user-attachments/assets/6b797fcb-1071-4c64-a60f-1ffca35931d2">

PR #1527, opened 2024-08-29
fix: add pnpm lock option

randallt21:
Fixes the issue here https://github.com/oclif/oclif/issues/1526

I have tested the fix for building and packing the tarball in my local environment.

We do not use a `pretarball` script, but hopefully that helps someone in the future. Tested it with a simple script and got no errors.
Screenshot 2024-08-29 at 4 52 23 PM

### DIFF
--- a/src/tarballs/build.ts
+++ b/src/tarballs/build.ts
@@ -174,6 +174,9 @@ const addDependencies = async (c: BuildConfig) => {
         throw error
       }
     }
+  } else if (existsSync(path.join(c.root, 'pnpm-lock.yaml'))) {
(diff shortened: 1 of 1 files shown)


mdonnalley (approved):

mdonnalley:
Thanks @randallt21 for the contribution 🏆 

I've merged this into a separate branch so that the integration tests can run (they don't run on forks). I'll merge that PR to main as soon as the tests pass

randallt21:
Sounds good @mdonnalley! 

Thank you for being on top of it and getting this through.

PR #1577, opened 2024-10-15
Add explicit error when `files` is missing from package.json

mathieudutour:
If `files` is missing from the package.json, `oclif manifest` would throw an unhelpful error

```
TypeError: Cannot read properties of undefined (reading 'find')
```

This makes sure that the user knows how to fix the issue

### DIFF
--- a/src/commands/manifest.ts
+++ b/src/commands/manifest.ts
@@ -96,6 +96,14 @@ export default class Manifest extends Command {
       await plugin.load()
     }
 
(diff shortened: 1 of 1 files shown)


mdonnalley:
@mathieudutour Thanks for the contribution 🏆 

Since we don't run integration tests on forks, I'm going to merge this into a separate feature branch so that they can run. Once they pass, I'll merge to main

PR #1578, opened 2024-10-15
fix: add explicit error when `files` is missing from package.json (#1577)

mdonnalley:
If `files` is missing from the package.json, `oclif manifest` would throw an unhelpful error

```
TypeError: Cannot read properties of undefined (reading 'find')
```

This makes sure that the user knows how to fix the issue


iowillhoit (approved):
LG

PR #1585, opened 2024-10-21
set node versions for testing to 20 and 22

mdonnalley:
Using `node_version: [lts/*, latest]` brought in node 23 which apparently isn't playing nicely with aws-sdk. So for now hardcode to 20 and 22


iowillhoit (approved):
LG

PR #1618, opened 2024-11-29
Feat: allow for a custom tsconfig path during readme generation

SpanishPear:
👋  I'd like the ability to point the tsconfig file at a different filename when running generate-readme e.g. `my-tsconfig.json`

I've added a flag following the same format and description as the README path.

Thanks!


SpanishPear:
Signed

PR #1673, opened 2025-01-22
fix: use eslint 9

### LINKED ISSUE #1665: oclif generate should use eslint 9 compatible templates
**Is your feature request related to a problem? Please describe.**

Running the generate command to create a new oclif cli creates a project with a dependency on a version of eslint that is EOL since October 2024 (8.57.1).  Upgrading to eslint 9 is non-trivial for generated projects because the a new eslint config format is required as of 9.x and many of the oclif eslint projects are not compatible with the new config format. 

**Describe the solution you'd like**
I know this is no fun seemingly low value maintenance but running npm install on newly generated cli projects now throws many warnings due to outdated packages mostly related to these eslint rule projects:

```plaintext
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated @humanwhocodes/config-array@0.13.0: Use @eslint/config-array instead
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@8.1.0: Glob versions prior to v9 are no longer supported
npm warn deprecated @humanwhocodes/object-schema@2.0.3: Use @eslint/object-schema instead
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated eslint@8.57.1: This version is no longer supported. Please see https://eslint.org/version-support for other options.
```

**Describe alternatives you've considered**
I have done the work of creating what I think to be a config file compatible with oclif standard aslant rules from `.eslintrc.json`. 

```json
{
  "extends": ["oclif", "oclif-typescript", "prettier"]
}
```

The result is shockingly complex 100+ lines of configuration because these other eslint rule projects are also using incompatible formats. I am happy to share it, but a better approach would be update the dependencies or introduce a new ruleset project that is eslint 9 compatible.

**Why do I care**
We are nearing release on an open source Oclif based system for building ai agents where the a key extension mechanism is oclif's plugin system. We believe that this will drive many new oclif plugins and we'd like to both maintain oclif conventions rather than introducing our own but also don't want users to have the experience of feeling like their plugins are using out-of-date, unsupported dependencies right out of the gate.

mdonnalley:
Use eslint 9 config

Depends on https://github.com/oclif/eslint-config-oclif/pull/122

Fixes #1665
@W-17599563@


cristiand391 (approved):

cristiand391:
**QA notes**:

setup:
generated an ESM/CJS plugin with `./bin/run.js generate test-eslint`, modified `hello world` command like this:
```ts
import {Command} from '@oclif/core'
import { EOL } from 'os'

export default class World extends Command {
  static args = {}
  static description = 'Say hello world'
  static examples = [
    `<%= config.bin %> <%= command.id %>
hello world! (./src/commands/hello/world.ts)
`,
  ]
  static flags = {}

  async run(): Promise<void> {
    console.log(EOL)

    this.log('hello world! (./src/commands/hello/world.ts)')
  }
}
```

✅ CJS/ESM plugins work with default eslint v9 config (`yarn lint`)

❌ missing `prefer-node-protocol` warning, the eslint-config-oclif PR is missing the unicorn/recommended rules: 
https://github.com/oclif/eslint-config-oclif/blob/eb27b1ca77fee27253ff341d6dd926b1e7ae3801/index.js#L6

✅ eslint-plugin-mocha rules apply in tests

✅ `no-unused-expressions` disabled on test files
❓ : `dangle-comma` was deprecated on v8:
https://eslint.org/docs/latest/rules/comma-dangle

if we don't want to bring in another `stylistic` stuff maybe just remove it or see if prettier handles this?

ditto:
`no-multi-spaces`: https://eslint.org/docs/latest/rules/no-multi-spaces#rule-details
`indent`: https://eslint.org/docs/latest/rules/indent#rule-details
`quotes`: https://eslint.org/docs/latest/rules/quotes
`semi`: https://eslint.org/docs/latest/rules/semi

iowillhoit (approved):
LG

PR #1679, opened 2025-01-29
feat: ensure jit plugins are downloaded via npm

tristanzander:
This prevents errors that occur due to tarballs needing to be downloaded with
authentication. Using NPM's authentication schemes makes it simple and easy to include JIT plugins
from within private organizations.

### Context

I was following the guide to implement JIT plugins using your guide posted [here](https://oclif.io/docs/jit_plugins) when I came across the following error:
```bash
# npx oclif manifest
Generating JIT plugin manifests... !
    HTTPError: Response code 401 (Unauthorized)
    Code: ERR_NON_2XX_3XX_RESPONSE
```

This occurs because oclif tries to download the tarball directly from the source without going through npm. I had originally thought to include a `authentication` key to the `oclif.jitPlugins['pluginName']` object within the oclif config, but it ended up being much easier to use `npm pack` directly to retrieve the tarball. Hopefully, this implementation supports downloading every type of package that can be installed through npm, which should cover every type of internal NPM repository.

The tests aren't as thorough as I would like, but I'm not totally sure how to test it better. Please let me know if you have any suggestions, or whether you think this is enough.


tristanzander:
Closing an re-opening due to CLA

PR #1698, opened 2025-02-15
chore: add cli keyword

Naymi:
I found this package by chance, only after I looked at the first few pages on the NPM with the key word CLI
Perhaps it is worth adding this keyword to find this package easier?


Naymi:
I can't sign the Salesforce Inc. Contributor License Agreement. The bot doesn't work 🥲

mdonnalley (approved):

Issue #1764, opened 2025-05-23
`upload tarballs` fails with S3-compatible storage due to AWS SDK v3.729.0+ mandatory checksums


toxsick:
### Environment
- **oclif version**: 4.17.45
- **Node.js version**: 22.15.1
- **Platform**: macOS
- **Storage**: QNAP S3-compatible storage (QuObjects)

### Description

`oclif upload tarballs` command fails when uploading to S3-compatible storage systems (QNAP, Cloudflare R2, MinIO, etc.) due to mandatory checksum headers introduced in AWS SDK v3.729.0+ that these systems don't support.

### Steps to Reproduce

1. Configure oclif with S3-compatible storage endpoint (non-AWS)
2. Run `oclif upload tarballs`
3. Upload fails with streaming request error

### Expected Behavior

Tarballs should upload successfully to S3-compatible storage, as they did with earlier oclif versions.

### Actual Behavior

Upload fails with the following error:

```
An error was encountered in a non-retryable streaming request.
    499: UnknownError
    at throwDefaultError (~/node_modules/@smithy/smithy-client/dist-cjs/index.js:867:20)
    at ~/node_modules/@aws-sdk/client-s3/dist-cjs/index.js:4952:14
    at async ~/node_modules/@smithy/middleware-serde/dist-cjs/index.js:36:20
    at async ~/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:484:18
    at async ~/node_modules/@smithy/middleware-retry/dist-cjs/index.js:320:38
```

### Root Cause

AWS SDK for JavaScript v3.729.0 introduced mandatory checksum algorithms for S3 uploads (`x-amz-checksum-crc32`, etc.) that S3-compatible storage systems don't implement, causing them to return errors.

**Timeline:**
- ✅ **Working**: `@aws-sdk/client-s3@3.726.1` and earlier
- ❌ **Broken**: `@aws-sdk/client-s3@3.729.0` and later
- 📍 **Current in oclif**: `@aws-sdk/client-s3@3.812.0`

### References

This is a known issue affecting all S3-compatible storage:
- [Cloudflare R2 AWS SDK Compatibility](https://developers.cloudflare.com/r2/examples/aws/aws-sdk-js-v3/)
- [Community Report](https://community.cloudflare.com/t/aws-sdk-client-s3-v3-729-0-breaks-uploadpart-and-putobject-r2-s3-api-compatibility/758637)
[...]

PR #1782, opened 2025-06-16
feat: support S3 compatible storage types

chaitanyya:
Adds support to skip checksum for S3 compatible storage like Cloudflare R2 that are currently failing.

Also requested here: https://github.com/oclif/oclif/issues/1764


chaitanyya:
i just signed the CLA but it's still saying it's not signed :/

Issue #1812, opened 2025-07-17
Customizable SHA hash for pack/upload


otaviojacobi:
Currently, oclif `pack` and `upload` uses the checked out commit hash to upload to s3. That present limitations for CI/CD workflows that might detach the commit head temporarily. I would be interested to know if you are open for a PR adding an optional `--sha` flag (similar to the promote one) to be a 7 digit value (same format as sha) so that workflows on the same commit but checked out could override it?

PR #1813, opened 2025-07-17
feat: support overriding sha on pack & upload

otaviojacobi:
Allows overriding the `sha` hash used by `pack` & `upload` commands.
See: https://github.com/oclif/oclif/issues/1812

This is a first version and I am up for any changes needed. Specially, if there are any specific test cases we want to run for this flag I would be happy to do them aswell.


mdonnalley (approved):
Changes look good to me - thanks for the PR! I'll merge into a separate branch so integration tests can run. Once those pass, I'll merge to main

PR #1852, opened 2025-08-28
Wr/win installer cmd exe

WillieRuemmele:
@W-19315214@

fixes the default nsis installer to use the global cmd.exe - rather than a possibly malicious, local cmd.exe

### DIFF
--- a/src/commands/pack/win.ts
+++ b/src/commands/pack/win.ts
@@ -75,16 +75,28 @@ Section "${config.name} CLI \${VERSION}"
   SetOutPath $INSTDIR
   File /r bin
   File /r client
-  ; Use explicit System32/Sysnative path to cmd.exe for security
-  ; Check if we're running on x64
-  System::Call "kernel32::GetCurrentProcess() i .s"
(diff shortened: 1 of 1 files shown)


src/commands/pack/win.ts
```diff
@@ -74,7 +74,16 @@ Section "${config.name} CLI \${VERSION}"
...
+
+  ; Use explicit System32/Sysnative path to cmd.exe for security
+  StrCpy $0 "$WINDIR\\System32\\cmd.exe"  ; Try System32 first
+  IfFileExists "$0" path_is_safe
+    StrCpy $0 "$WINDIR\\Sysnative\\cmd.exe"  ; Try Sysnative for WOW64
+    IfFileExists "$0" path_is_safe
+      MessageBox MB_OK|MB_ICONSTOP "Error: Could not find system cmd.exe. Installation cannot continue."
+      Abort
+      
+  path_is_safe:
```
shetzel, line 86: Are you sure these are the paths for all Windows systems?  I'd hate to break customers that have a valid cmd.exe path that are not one of these two.

shetzel (approved):

PR #2008, opened 2026-03-24
feat(readme): allow disabling source links

rexxars:
For some CLI's, showing links to source locations isn't actually helpful/wanted. This PR adds a new `--[no-]source-links` flag, that lets you disable them should you not want to.

Wasn't sure if I should add a setting to `package.json` or not, but went for the minimal feature set first.


jfeingold35 (approved):

jfeingold35:
@rexxars , okay, it looks like the failures are reproducible in `main`, so it's not something you've introduced. We'll get them fixed ASAP in order to unblock you.

PR #2023, opened 2026-05-13
fix: bumping oclif/core to 4.11.2 @W-22477677@

jfeingold35:
As part of @W-22477677@, the dependency on oclif/core must be updated to the latest version.
This also unpins a number of dependencies that seem to have been wrongfully pinned when they were updated in #1992 .


package.json
```diff
@@ -8,12 +8,12 @@
   },
   "bugs": "https://github.com/oclif/oclif/issues",
   "dependencies": {
-    "@aws-sdk/client-cloudfront": "3.1009.0",
-    "@aws-sdk/client-s3": "3.1014.0",
+    "@aws-sdk/client-cloudfront": "^3.1009.0",
```
jfeingold35, line 11: Way back in #1992 , I merged a PR that was a consolidation of a bunch of smaller Dependabot PRs. I didn't catch it at the time, but that PR actually pinned all of those dependencies (including `@oclif/core`), despite them not previously being pinned. As far as I can tell, there's no good reason for them to be pinned, so I'm unpinning them.
This is only _truly_ important for `@oclif/core`, as it allows the dependency de-duping to start working once again, but I figured it was worth doing for all of them.

.github/workflows/test.yml
```diff
@@ -112,7 +112,8 @@ jobs:
           node-version: ${{ matrix.node_version }}
           cache: yarn
       - if: matrix.package_manager == 'pnpm'
-        run: 'npm install -g pnpm'
+        # TODO [JAMIE]: ADD SUPPORT FOR PNPM v11!
```
jfeingold35: PNPM released v11 within the last month or so, and it makes several major changes from v11. Most notably:
- whereas it was previously ignoring post-install scripts with a warning, now it throws an error.
- it no longer reads any properties from the `package.json`'s `pnpm` property, instead looking for a `pnpm-workspace.yml` file at the root of the workspace.

Collectively, those mean that we cannot in our current state support PNPM v11. So I'm just hardcoding it to use v10 for now, and we can add a tech debt work item for v11 support along with a (possibly pinned) Issue in this repo so that people know we're already aware of it.

iowillhoit (approved):

PR #2040, opened 2026-05-31
fix: support CRLF readme tags

### LINKED ISSUE #793: oclif readme only searches for LF in regex, not CRLF, which makes it incompatible on windows
https://github.com/oclif/oclif/blob/58cefb3be1437ede9885d8a415951179eed08548/src/commands/readme.ts#L75

Expected result after running npx oclif readme is that all 3 tags are replaced.
Actual result on windows with CRLF:
```
<!-- usage -->
sh-session
$ npm install -g package
$ package COMMAND
running command...
$ package(--version)
package/1.2.6 win32-x64 node-v16.13.1
$ package --help [COMMAND]
USAGE
  $ package COMMAND
...

<!-- usagestop -->
sh-session
$ npm install -g package
$ package COMMAND
running command...
$ package (--version)
package/1.2.6 win32-x64 node-v16.13.1
$ package--help [COMMAND]
USAGE
  $ package COMMAND
...

<!-- usagestop -->
```
This is because the replace command dosent replace the tags and the body in the tags with `<!-- usage -->`. So when line 78 in readme.ts is executed it replaces the first tag and the rest of the body and the stop tag is left.

Possible mitigations: replace all occurences with \r\n with \n in commands/readme.ts. (readme = readme.replaceAll('\r\n', '\n')) or extends the regex to match \r\n as well.

Possible workarounds: 
1. Just use LF in vs code
2. Make a script to replace CRLF with LF before running oclif readme.
This can then be inserted into the prepack command in package.json. For example: "prepack": "yarn build && oclif manifest && node prepreadme.js && oclif readme"
prepreadme.js: 
```
const fs = require('fs')
const path = require('path')

const cwd = process.cwd()
const readmePath = path.resolve(cwd, 'README.md')

fs.readFile(readmePath, 'utf8', (err, readme) => {
  readme = readme.replaceAll('\r\n', '\n')
  fs.writeFile(readmePath, readme, () => console.log('replace \\r\\n with \\n readme.md'))
})
```
Otherwise oclif works well on windows ^^

cyphercodes:
## Summary
- Update README tag replacement to match sections with CRLF line endings.
- Add a unit test covering CRLF-delimited README usage tags.

Fixes #793

## Testing
- `yarn mocha --forbid-only test/unit/readme-generator.test.ts`
- `yarn build`
- `yarn test` (passes; lint reports existing TODO warnings only)

### DIFF
--- a/src/readme-generator.ts
+++ b/src/readme-generator.ts
@@ -209,7 +209,7 @@ export default class ReadmeGenerator {
   protected replaceTag(readme: string, tag: string, body: string): string {
     if (readme.includes(`<!-- ${tag} -->`)) {
       if (readme.includes(`<!-- ${tag}stop -->`)) {
-        readme = readme.replace(new RegExp(`<!-- ${tag} -->(.|\n)*<!-- ${tag}stop -->`, 'm'), `<!-- ${tag} -->`)
+        readme = readme.replace(new RegExp(`<!-- ${tag} -->[\\s\\S]*?<!-- ${tag}stop -->`, 'm'), `<!-- ${tag} -->`)
       }
 
       ux.stdout(`replacing <!-- ${tag} --> in ${this.options.readmePath}`)
--- a/test/unit/readme-generator.test.ts
+++ b/test/unit/readme-generator.test.ts
@@ -263,6 +263,19 @@ Title of Doc
     })
   })
 
+  describe('replaceTag', () => {
+    it('should replace tags when the README uses CRLF line endings', () => {
+      const generator = new TestReadmeGenerator(config, {outputDir: 'docs', readmePath: 'README.md'})
+      // @ts-expect-error because protected method
+      const rendered = generator.replaceTag(
+        '# Usage\r\n\r\n<!-- usage -->\r\nold usage\r\n<!-- usagestop -->\r\n\r\n# Commands\r\n',
+        'usage',
+        'new usage',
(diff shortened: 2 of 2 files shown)


jfeingold35:
@cyphercodes , can you please sign the CLA so that we can get this merged? Thanks!

jfeingold35:
@cyphercodes , also, there are some test failures that you should look into.

jfeingold35:
@cyphercodes , actually, these failures are reproducible in `main` and therefore unrelated to anything you've done. I'll be looking into this immediately and I'll keep you posted on whether there's anything you'll need to do. Most likely, you'll need to rebase your PR off of `main`.

jfeingold35:
@cyphercodes , if you could please rebase this PR off of `main`, and sign the SF CLA, then I can get this PR approved and merged. Thank you!

cyphercodes:
Rebased this PR onto current `main` as requested.

Local verification:
- `yarn install --immutable` ✅
- `yarn mocha test/unit/readme-generator.test.ts --forbid-only --timeout 600000` ✅
- `yarn eslint src/readme-generator.ts test/unit/readme-generator.test.ts` ✅
- `yarn tsc --noEmit` ✅
- `git diff --check` ✅

Note: the Salesforce CLA signature is still a manual/legal account action for @cyphercodes.

PR #2052, opened 2026-06-18
fix: prevent command injection by jit plugins @W-23028384@

### LINKED ISSUE #2051: Command injection in `oclif manifest` via `package.json` `oclif.jitPlugins`
**Is this a bug with the oclif CLI? Or the oclif framework?**

This appears to be a bug in the `oclif` CLI.

**Describe the bug**

`oclif manifest` reads `oclif.jitPlugins` from a target project's `package.json`. When JIT plugins are enabled, each plugin name and version is interpolated into an `npm pack` shell command:

```js
npm pack ${plugin}@${version} --pack-destination "${tarballStoragePath}" --json
```

That command string is then executed with `child_process.exec()`.

Because the `jitPlugins` key and value are not shell-escaped or validated before interpolation, shell metacharacters in a plugin name can execute arbitrary commands when `oclif manifest` is run against a crafted project.

This is reachable through the public CLI:

```bash
oclif manifest /path/to/project
```

**To Reproduce**

1. Create a test project with a crafted `oclif.jitPlugins` entry:

```bash
rm -rf /tmp/oclif-poc /tmp/oclif-aci
mkdir -p /tmp/oclif-poc

cat > /tmp/oclif-poc/package.json <<'JSON'
{
  "name": "oclif-jit-poc",
  "version": "1.0.0",
  "oclif": {
    "jitPlugins": {
      "lodash; touch /tmp/oclif-aci #": "1.0.0"
    }
  },
  "files": ["lib"]
}
JSON
```

2. Run `oclif manifest` against that project:

```bash
npx -y oclif@4.23.16 manifest /tmp/oclif-poc
```

3. Check whether the injected command ran:

```bash
test -f /tmp/oclif-aci && echo "VULNERABLE"
```

Observed output from my local reproduction:

```text
EXIT=1
MARKER=yes
Generating JIT plugin manifests... !
    SyntaxError: Unexpected token 'l', "lodash-4.18.1.tgz
    " is not valid JSON
```

The command exits later while parsing `npm pack` output, but `/tmp/oclif-aci` is created first, showing that the injected command already executed.

**Expected behavior**

Values from `package.json` should be treated only as npm package names and versions. Shell metacharacters such as `;`, `#`, `$()`, backticks, quotes, or spaces should not be interpreted by a shell.

`oclif manifest` should either:

- use `child_process.execFile()` or `spawn()` with `shell: false` and an argv array, or
- validate `jitPlugins` names and versions before invoking npm.

For example:

```js
execFile('npm', [
  'pack',
  `${plugin}@${version}`,
  '--pack-destination',
  tarballStoragePath,
  '--json',
], callback);
```

**Your Environment (please complete the following information):**

- `oclif` Version: `4.23.16`
- OS & version: macOS arm64
- Shell/terminal & version: zsh
- Node.js version: `v24.10.0`

**Additional context**

The vulnerable code path in the published `oclif@4.23.16` package is in `lib/commands/manifest.js`:

```js
async downloadTarball(plugin, version, tarballStoragePath) {
  const { stderr } = await this.executeCommand(
    `npm pack ${plugin}@${version} --pack-destination "${tarballStoragePath}" --json`
  );
  // ...
}

async executeCommand(command, options) {
  return new Promise((resolve) => {
    exec(command, options, (error, stderr, stdout) => {
      // ...
    });
  });
}
```

The data flow is:

```text
package.json oclif.jitPlugins key/value
  -> Object.entries(packageJson.oclif.jitPlugins)
  -> downloadTarball(jitPlugin, version, fullPath)
  -> npm pack ${plugin}@${version}
  -> child_process.exec(command)
```

Impact: if a developer, CI job, release bot, or automation pipeline runs `oclif manifest` on an untrusted or attacker-controlled project, arbitrary commands can execute with the privileges of that process.

jfeingold35:
Closes #2051 and @W-23028384@

Closes a security vulnerability wherein a maliciious JIT plugin entry could inject commands that would be executed.


src/commands/manifest.ts
```diff
@@ -106,12 +106,16 @@ export default class Manifest extends Command {
   }
 
   private async downloadTarball(plugin: string, version: string, tarballStoragePath: string): Promise<string> {
-    const {stderr} = await this.executeCommand(
-      `npm pack ${plugin}@${version} --pack-destination "${tarballStoragePath}" --json`,
-    )
+    const {stdout} = await this.executeCommand('npm', [
```
jfeingold35, line 110: This should have been `stdout` all along. It was only working in the first place due to a quirk in `exec` that doesn't exist with `spawn`.

src/commands/manifest.ts
```diff
@@ -124,17 +128,42 @@ export default class Manifest extends Command {
     return path.join(tarballStoragePath, filename)
   }
 
-  private async executeCommand(command: string, options?: ExecOptions): Promise<{stderr: string; stdout: string}> {
-    return new Promise((resolve) => {
-      exec(command, options, (error, stderr, stdout) => {
```
jfeingold35, line 129: This is what allowed us to use `stderr` earlier. In reality, `exec`'s streams are provided with `stdout` first and `stderr` second. So we actually had the streams mixed up here. `spawn` requires the streams to be tracked separately.

WillieRuemmele:
## Local Verification

Verified the fix locally using the steps to reproduce from #2051:

### 1. Regression test (new test fails without the fix)

Reverted `src/commands/manifest.ts` to the `main` branch version and ran the new unit test:

```
manifest
  ✔ should generate manifest (336ms)
  ✔ should generate jit plugins (838ms)
  1) is not vulnerable to command injection via JIT plugins

1 failing

  AssertionError: expected true to equal false
```

The `ATTACK_SUCCEEDED.txt` marker file was created, confirming the injection succeeds with the old `exec()` code.

### 2. PoC from issue #2051 (injection blocked with the fix)

Created the exact PoC project from the issue:

```json
{
  "name": "oclif-jit-poc",
  "oclif": {
    "jitPlugins": {
      "lodash; touch /tmp/oclif-aci #": "1.0.0"
    }
  },
  "files": ["lib"]
}
```

Ran `./bin/run.js manifest /tmp/oclif-poc` against it with the fix in place. Result:

```
npm error path .../lodash; touch /tmp/oclif-aci #@1.0.0/package.json
npm error enoent Could not read package.json: ENOENT
```

The malicious string was passed as a **literal argument** to npm (no shell interpretation). The injected `touch /tmp/oclif-aci` never executed:

```
$ test -f /tmp/oclif-aci && echo "VULNERABLE" || echo "NOT VULNERABLE"
NOT VULNERABLE
```

### Summary

- `cross-spawn` with an argv array eliminates shell metacharacter interpretation
- The `stderr` → `stdout` change for `npm pack --json` output is also correct (the old code had the `exec` callback params swapped: `(error, stderr, stdout)` instead of `(error, stdout, stderr)`)

WillieRuemmele (approved):

PR #2059, opened 2026-06-24
Update command to use exec for node process

Mika56:
Currently, the node process is launched directly. However, if the process is started by a process manager such as systemd, killing the main process (which is, in fact, a bash process and not the node process) will do nothing: bash will orphan the node process and leave it running.

By using exec, we're substituting the bash process with the node one: there is no bash process to kill anymore, only the node process directly.


jfeingold35:
@Mika56 , thanks for opening this. We'll get it reviewed as soon as possible, and we'll let you know if anything needs to change.

PR #2090, opened 2026-09-01
build: bumping ts6 (W-23983563)

jfeingold35:
Per @W-23983563@, we're bumping the dependency on TS from v5 to v6, as well as a few miscellaneous dependencies that we tweaked in previous TS6-related stories.


test/fixtures/cli-with-nested-topics/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "declaration": true,
     "importHelpers": true,
-    "module": "commonjs",
+    "module": "nodenext",
```
jfeingold35, line 5: Had to bump these to `nodenext` in order to get the tests to pass.

k80bowman (approved):

